PR #47, opened 2017-07-07
Pe crash

alvarofe:
I found this crash (NULL dereference) while fuzzing LIEF. I attach a possible fix and the sample.


romainthomas:
Hi @alvarofe, thank you for the bug report. I suggest this fix: https://github.com/lief-project/LIEF/commit/e16d9bda564cfb9342b7692b642ea1e7ee4afc9a because the check is done at the creation of the ``ResourcesManager`` object and not in its methods.

romainthomas:
Merged with https://github.com/lief-project/LIEF/commit/7006c905c583f856774b2b714916fe8c8ccbd98b

alvarofe:
It makes more sense in the creation of the class. It was a rather quick fix :)

PR #75, opened 2017-08-01
Fix MachO header flag API usage after cbe8354

jevinskie:
Without this I cannot build on Mac.


romainthomas:
Fix with https://github.com/lief-project/LIEF/commit/2a9b156f999022e08d4ba019e755b908ceec1176

PR #116, opened 2017-11-19
Reorg code so LIEF_LOGGING=off works properly

illera88:
Before this comit even when `LIEF_LOGGING=off` [easyloggingpp](https://github.com/muflihun/easyloggingpp) is still compiled and some of its functions included in `libLIEF`.


illera88:
False negative with travis.

romainthomas:
Everything is green ;)

PR #128, opened 2017-12-14
Parse the MachO LC_RPATH command

0xbf00:
This patch adds parsing for the LC_RPATH command. The LC_RPATH command is used to specify an additional root path where libraries can be found. Multiple LC_RPATH commands can be included in a single MachO file.

Unfortunately, none of the test binaries feature such a command. Therefore, I have not yet added a test, though I did test the functionality on my own.


romainthomas:
Very nice! I'll add tests as well as a binary having this command before merging

0xbf00:
I'd gladly take this off your plate. I have a small test binary (~12 K) already and can write a test later today. How does this sound to you?

0xbf00:
I've now added a small test case. Automated tests will likely fail right now, because my sample has not been merged into the samples repository.

romainthomas:
I merged the test case and restarted the CI

romainthomas:
Thanks for the PR

PR #160, opened 2018-05-02
fixed inverted ENDIAN_LITTLE and _BIG values

yd0b0N:
Simple fix for inverted values between big and little endian designations in ELF headers

### DIFF
--- a/include/LIEF/Abstract/enums.hpp
+++ b/include/LIEF/Abstract/enums.hpp
@@ -69,8 +69,8 @@ enum MODES {
 
 enum ENDIANNESS {
   ENDIAN_NONE   = 0,
-  ENDIAN_BIG    = 1,
-  ENDIAN_LITTLE = 2,
+  ENDIAN_LITTLE = 1,
+  ENDIAN_BIG    = 2,
 };
 
 


romainthomas:
Why changing these values ?

yd0b0N:
Hi Romain,
Well I based the change on the [ELF specification](http://refspecs.linuxbase.org/elf/elf.pdf) that contains the following in "ELF Identification/EI_DATA" (page 1-7):
image
Wherein it looks as if LSB/Little Endian uses value 1 and MSB/Big Endian uses value 2. The pre-commit values were inverted in LIEF (and were preventing my endian-aware version from properly interpreting PPC binaries).

romainthomas:
Yes I agree but this part is not related to ELF format. Thus I think this change is not necessary 

https://github.com/lief-project/LIEF/blob/55e99450f4b2bc069b618891b4e1ad3504946a0a/src/ELF/Header.cpp#L52-L56

https://github.com/lief-project/LIEF/blob/55e99450f4b2bc069b618891b4e1ad3504946a0a/src/MachO/Header.cpp#L51-L60

yd0b0N:
Got it, my bad: I hadn't noticed that the ENDIANNESS enum wasn't ELF specific. I fixed up my code to use ELF_DATA instead of ENDIANNESS where relevant.
Sorry about that...

PR #161, opened 2018-05-02
Ppc relocs

yd0b0N:
Support for PPC and PPC64 relocations.
Note this patch in itself isn't sufficient for complete PPC support because it does not address manipulation of foreign-endian data. This latter aspect still needs to be hammered out.


romainthomas:
Merged with https://github.com/lief-project/LIEF/commit/08b514191f661eeabbdf8ecacd1d7dd35a67ca54

Thanks!

PR #162, opened 2018-05-02
Add applies_to field to Relocation

yd0b0N:
This patch adds an `applies_to` field to the `ELF.Relocation` object that tracks with `ELF.Section` the relocation applies to in an ELF _object_ file. It's value is provided by the `ELF.Parser` when the latter analyzes an ELF object file. When dealing with object files, the field is necessary to properly interpret the meaning of the relocation's `address` field because in this context, instead of representing a virtual address, the address is relative to the beginning of the designated ELF section.


romainthomas:
Merged with https://github.com/lief-project/LIEF/commit/0bbeaa81e9fdc95f3ed19cc1b8469d45c6bca660.

I did small modifications, especially I replaced the name ``applies_to`` with ``section`` because the relocation could also apply to a segment, a symbol etc

Thanks for the contribution.

PR #166, opened 2018-05-08
Endian support for ELF parsing

yd0b0N:
This pull requests allows lief to _parse_ ELF binaries with a foreign-endian encoding. It also adds a little fix to how Note sections in ELFs are parsed (read in as 4 byte words) so that foreign-endian instances retain the expected values.


yd0b0N:
Just noticed that there is a regression when reading native-endian Note sections (at least on x86-64). So hold off on merging this until that issue gets sorted out.

yd0b0N:
Should be ok now. My sample ELFs, both BE and LE are being parsed fine.

romainthomas:
Hum, CI still fail.

yd0b0N:
I figured out how to get the unit tests running on my machine and observed two problems:
- Failed **ELF_PYTHON_test_dynamic**: as I understand things this one involves renaming a dynamic library and altering a binary to use the new name. I couldn't really see where my modifications could have an impact, so I just checked out the upstream branch (commit id `e7c45a1`) and try to run the test against it. As it turns out the test failed against this branch too (at least when run on my machine). So there might be an issue lurking elsewhere.
- Failed **FUZZER_ELF_bin_ls_x86** & **FUZZER_ELF_bin_nm_x86**: these are Melkor fuzz tests that are indeed failing. I still need to debug one of the crash-inducing samples to see if I can figure out what was broken in _pylief.so.

romainthomas:
Looks good :) I'll do a deep review next week, Thanks!

romainthomas:
I almost finished the review. It will be merged by the end of the day.

romainthomas:
Merged with https://github.com/lief-project/LIEF/commit/e794ac1502ee7636755bd441923368f88525a7d0

PR #167, opened 2018-05-09
Fix possible out of bound reads in utils.cpp

aurelianito:
Use the newly created stream methods to check out of bounds in `PE/utils.cpp`


romainthomas:
Thanks!

PR #183, opened 2018-06-15
exception-free headers

serge-sans-paille:
With the extra bonus of internalizing some headers that should not be public, or so I think :-)


romainthomas:
Thank you @serge-sans-paille

PR #189, opened 2018-07-05
allow for reproducible builds by zeroing non-initialized dosheader fields

jbremer:
Multiple runs of the same code would result in binaries with different hashes, this PR fixes that issue.


romainthomas:
Merged. Thanks!

PR #201, opened 2018-07-31
Remove double building of static+shared library, and fix the hardcoded MSVCRT version

aguinet:
I think it needs some fixes at the CI level, that I didn't take the time to do!


romainthomas:
Merged!

aguinet:
Thanks Romain :)

PR #212, opened 2018-09-13
api/c: Add explict size fields for extracted content fields

hillu:
Corrupted files may cause section or segment contents not to be
available entirely. However, C-based client code is not able to detect
this which may cause out-of-bounds reads when accessing data from the
.content fields.

Note: This is a silent ABI-breaking change because it changes struct layouts.

Note: Naming is inconsistent: .size was already taken in PE Sections.


romainthomas:
Thanks!

PR #222, opened 2018-10-19
Fix/export trie

pbrunet:


src/MachO/Builder.cpp
```diff
@@ -333,7 +333,8 @@ void Builder::write(const std::string& filename) const {
         std::begin(content),
         std::end(content),
         std::ostreambuf_iterator<char>(output_file));
-  }
+  } else
+    std::cerr << "Fail to write binary file\n";
```
romainthomas: You can use the logger module: 
```cpp
LOG(ERROR) << "Fail to write binry file";
```

Issue #245, opened 2018-12-17
PE resources parsed incorrectly


williballenthin:
I've noticed that LIEF occasionally parses PE resources incorrectly; specifically, string items from the version info.

As an example, consider `71b201a5a7dfdbe91c0a7783f845b71d066c62014b944f488de5aec6272f907c`, which is available via VT (let me know if I should send this directly).

Resource Hacker parses the following structures:

![image](https://user-images.githubusercontent.com/156560/50116160-a0a8ff00-0206-11e9-8360-b2b98ecd506e.png)

Note that there are values like: `CompanyName`=`Microsoft Corporation. All rights reserved.`

However, LIEF parses values differently:

![image](https://user-images.githubusercontent.com/156560/50116221-ce8e4380-0206-11e9-897d-924a24000897.png)

```
'.7600.16385': ' ',
'1, 7600, 16385': ' ',
'Comments': 'x,\x01CompanyName',
'ecialBuild': 'D',
'galTrademarks': '(',
'iginalFilename': ' ',
'ivateBuild': ' ',
'leDescription': '>\x0f\x01FileVersion',
'oductName': 'H\x12\x01ProductVersion',
'rosoft Corporation. All rights reserved.': '(',
'ternalName': 'J\x13\x01LegalCopyright',
'yright (C) 2014': '('
```

I will begin to look into the parsing; however, I'm not yet familiar with the LIEF codebase, so cannot promise that this will be efficient.

williballenthin:
I took some time to manually parse the version resource for this sample, and I believe it illustrates where the bug is.

![image](https://user-images.githubusercontent.com/156560/50119763-faaec200-0210-11e9-8462-339c2904d21a.png)

Starting after the `StringFileInfo` header, there is the `key` field (offset 0x80), and then the list of key-value pairs. The illustration above shows two key-value pairs parsed out: in yellow (offset 0x98), and in blue (0xB0).

I think LIEF may incorrectly parse the yellow key-value pair, probably because the value portion is unexpectedly empty. This can be determined in two ways:
  1. The `string length in words` field is 0x0, and 
  2. the current position after the key ("Comment") equals the total element size (0x18).

I think the bug can be fixed by adding additional checks around line [ResourceManager.cpp::605](https://github.com/lief-project/LIEF/blob/1f32b253295535de77f81997ac99c49fb8015c83/src/PE/ResourcesManager.cpp#L605) to:
  1. assert `string_value_length` is greater than zero, and
[...]

PR #246, opened 2018-12-17
PE ResourcesManager: correctly handle empty strings

williballenthin:
addresses https://github.com/lief-project/LIEF/issues/245

Implements the fixes described there. The empty string is now parsed correctly:

image

Note that I haven't inspected the `get_var_file_info` routine, which could have a similar issue.

### DIFF
--- a/src/PE/ResourcesManager.cpp
+++ b/src/PE/ResourcesManager.cpp
@@ -589,6 +589,8 @@ ResourceStringFileInfo ResourcesManager::get_string_file_info(const VectorStream
     // Parse 'String'
     // ==============
     while (stream.pos() < end_offset) {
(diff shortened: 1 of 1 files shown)


romainthomas:
Thank you @williballenthin !

williballenthin:
would you consider tagging 0.9.1 so I can install via pypi and include these fixes?

romainthomas:
A new version of LIEF will be tagged in few months. You can always install lastest commit with pip using:
```console
$ pip install https://github.com/lief-project/packages/raw/lief-master-latest/pylief-0.9.0.dev.zip
```

Also LIEF 0.10 is coming with a better interface to install Python binding

williballenthin:
nightly build works perfectly, thanks!

PR #257, opened 2019-01-10
resources manager: fix parsing string info by trusting k-v pair header

### LINKED ISSUE #256: PE resources parsed incorrectly (2)
I've noticed that LIEF occasionally parses PE resources incorrectly; specifically, string items from the version info.

As an example, consider c044a6c5dee1533b2db0b8b4042d14e8d2e122b7ab4d075ba58839d263633cf3, which is available via VT (let me know if I should send this directly).

Resource Hacker parses the following structures:

image

Note that there are values like: CompanyName=Microsoft Corporation. All rights reserved.

However, LIEF parses values differently:

image

```
                               '\x0e\x01OriginalFilename': '',
                               '\x12\x01FileDescription': '',
                               '"\x01LegalCopyright': '',
                               'CompanyName': 'Micro',
                               'InternalName': 'uishow',
                               'ProductVersion': '0.5',
                               'dor Computer ': '*\x05\x01FileVersion',
                               'howfx.exe': '4\n\x01ProductName',
                               'icrosoft Corporation. Al': ' r',
                               'roso': ''
```

williballenthin:
closes #256 

with this patch, the version info is parsed correctly:

image

i have *not* inspected other parts of the resource parser (e.g. var info) for similar bugs since i don't have a test case.

### DIFF
--- a/src/PE/ResourcesManager.cpp
+++ b/src/PE/ResourcesManager.cpp
@@ -637,6 +637,11 @@ ResourceStringFileInfo ResourcesManager::get_string_file_info(const VectorStream
         VLOG(VDEBUG) << "Value: (empty)";
       }
 
+      const size_t expected_end = string_offset + string_length;
+
+      if (stream.pos() < expected_end && expected_end < end_offset) {
+        stream.setpos(expected_end);
+      }
       stream.align(sizeof(uint32_t));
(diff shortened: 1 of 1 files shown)


romainthomas:
Thanks !

PR #275, opened 2019-03-11
Counter bugs in PE resources manipulation functions

DaLynX:
Functions check if parent node `has_name` instead of checking it on the child passed in parameter.


romainthomas:
Thanks

PR #286, opened 2019-03-29
Add support for adding LC_UNIXTHREAD commands in a MachO

nezetic:
Add support for adding LC_UNIXTHREAD commands in a MachO.

Also include:
- Fix ```Binary::remove``` (exception raised)
- Fix symbols' values when segments are shifted


romainthomas:
Thanks :)

PR #308, opened 2019-06-25
Fix performance issue parsing relocations.

jlewallen:
Uses a map to track added relocations.

I just took a quick stab at this because I was seeing performance issues. More than willing to do whatever is necessary to this PR to get it fixed.


romainthomas:
Thank you @jlewallen

PR #320, opened 2019-07-25
Expose the raw bytes backing each embedded PE Authenticode cert

recvfrom:
This makes it trivial to extract the embedded certificates, which is really useful


recvfrom:
Weird, it looks like the CI failed with error:
```
fatal: unable to access 'https://github.com/lief-project/samples.git/': Could not resolve host: github.com
```

romainthomas:
I restarted :)

PR #337, opened 2019-08-27
Build manylinux1 wheels

lkollar:
Build `manylinux1` wheels using the `manylinux1_x86_64` Docker image. The Linux builds is executed in Docker and the produced `.whl` files are passed to `auditwheel repair` to correctly label them and vendor external dependencies inside the wheels.

A few notes on the changes:

- The Melkor ELF fuzzer is excluded from the build as its Makefile calls `find` with the `-executable` option which is not supported on CentOS 5. Given that the project seems fairly inactive and a PR for this might not get merged, I just patch the `Makefile` during the build process.
- I had to add a check to see if we have GLIBC 2.17 and skip a few tests which have been compiled  on a more recent platform as CentOS 5 does not have the required symbols. Either the tests should be recompiled on the `manylinux` Docker image or we could copy the resulting LIEF wheel out of the Docker container, install it on the host and run the test suite there.
- The builds seem slowerr in Docker. I've added a bind mount volume for the `ccache` data I'm not sure it works because I have not seen any speedup in build times.
- The Codacy check flags the `FROM` field in the `Dockerfile` for not using a specific tag. I think in this case it is actually preferable to not use a tag so that we always pick up the most recent build tools and the native library versions won't change in the image. Besides, the only tag on the image is `latest`. I don't know how to suppress the Codacy check though.

Resolves: #325

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,60 +24,31 @@ matrix:
         #    ]
 
 
-        # Linux gcc 5.0 - Python 3.5
-        - os: linux
-          dist: trusty
-          compiler: gcc
-          addons:
-            apt:
-              sources: ['ubuntu-toolchain-r-test', 'deadsnakes']
-              packages: ['g++-5', 'gcc-5', 'ccache', 'python3.5-dev']
+        - sudo: required
+          services:
+            - docker
           env: >
             [
-            CXX_COMPILER=g++-5                         ,
-            C_COMPILER=gcc-5                           ,
-            PYTHON_VERSION=3.5                         ,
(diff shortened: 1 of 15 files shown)


lkollar:
@romainthomas this is now ready for review.

romainthomas:
Thank you very much @lkollar!  I'll do the review this week-end

romainthomas:
It's great! Thank you. Would you mind to squash your commits ?

lkollar:
@romainthomas I squashed the commits.

PR #347, opened 2019-09-19
Use the right Python interpreter for Linux builds

lkollar:
The Dockerized .travis.yml builds attempt to invoke the interpreter in
the PYTHON_BINARY environment variable, which is only valid inside the
Docker image. To fix this, override the variable on Linux for tasks
which require the host's Python interpreter.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -169,6 +169,7 @@ script:
 
 after_success:
   - make package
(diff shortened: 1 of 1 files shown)


lkollar:
This should fix the issue but #348 is still causing silent errors.

romainthomas:
Looks like there is still an error

PR #351, opened 2019-09-25
Expose raw bytes of Authenticode-related structures

recvfrom:
Part of Authenticode verification consists of:
 - Comparing the computed Authenticode hash to the digest
   stored in the ContentInfo section
 - Comparing hash(ContentInfo) to the digest stored in the
   AuthenticatedAttributes section
 - Verifying signed(hash(AuthenticatedAttributes)) using a
   certificate identified by the issuer and serial number
   specified in the SignerInfo section

This commit makes it so that the raw bytes needed to
calculate hash(ContentInfo) and hash(AuthenticatedAttributes)
are available for use.

We could compute the hashes in LIEF and just make those
available instead, or we could perform the validation
outright and pass this info back through a boolean value
or something... Just exposing the raw bytes and letting
the application perform this check (if needed) is simpler,
though (especially for me, since I'm not familiar with a
lot of the C++ features used in this code base).

### DIFF
--- a/api/python/PE/objects/signature/pyAuthenticatedAttributes.cpp
+++ b/api/python/PE/objects/signature/pyAuthenticatedAttributes.cpp
@@ -58,6 +58,10 @@ void create<AuthenticatedAttributes>(py::module& m) {
         },
         "Return an URL to website with more information about the signer")
 
+    .def_property_readonly("raw",
+        &AuthenticatedAttributes::raw,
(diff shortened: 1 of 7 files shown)


recvfrom:
After doing some more testing it looks like it might be worth changing what's covered by authenticated_attributes.raw... I'll resubmit once I've done more testing.

PR #353, opened 2019-09-27
Remove .git directory from .dockerignore

lkollar:
As `setup.py` uses `git` to determine the version number, we need to copy the contents of `.git` into the image. This currently causes the `0.0.0` version to be used as the `get_version` function [fails to correctly determine](https://travis-ci.org/lief-project/LIEF/jobs/589846000#L596) the version.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,1 @@
 build
-.git


lkollar:
This seems to fix the issue: https://travis-ci.org/lief-project/LIEF/jobs/590317350#L597

lkollar:
Ping @romainthomas

romainthomas:
Thanks (I was on vacation)

PR #359, opened 2019-11-22
[PE] Fix missing offset in resource parsing

majin42:
This pull request fixes an issue that occurs while parsing the resources of
some Portable executables.
I believe it has been introduced while enhancing the BinaryStream interface:
https://github.com/lief-project/LIEF/commit/4ef839c3d8ed74663200531f73c7a5173ebce66c?diff=split#diff-e8bd49b1ac1ffa6531743048f188baefR380

As we can see in the commit above, the `offset` parameter is not passed anymore to the `can_read()` method
while the next operation is to retrieve the `data_entry` by peeking the stream at the previously computed offset.
This means that the parsing of a `pe_resource_data_entry` only occurs if the stream is already positioned
at `offset`.

Passing the offset parameter back fixes the issue.

### DIFF
--- a/src/PE/Parser.cpp
+++ b/src/PE/Parser.cpp
@@ -385,7 +385,7 @@ ResourceNode* Parser::parse_resource_node(
     if ((0x80000000 & data_rva) == 0) { // We are on a leaf
       uint32_t offset = base_offset + data_rva;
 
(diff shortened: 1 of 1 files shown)


romainthomas:
Thank you!

PR #365, opened 2019-12-09
Export BuildToolVersion and fix typo

mingwandroid:
Just a simple fix.. hope it's ok!

### DIFF
--- a/api/python/MachO/objects/pyBuildVersion.cpp
+++ b/api/python/MachO/objects/pyBuildVersion.cpp
@@ -50,7 +50,7 @@ void create<BuildVersion>(py::module& m) {
   tool_version_cls
     .def_property_readonly("tool",
         &BuildToolVersion::tool,
-        "" RST_CLASS_REF(.BuildeVersion.TOOLS) " type")
+        "" RST_CLASS_REF(.BuildVersion.TOOLS) " type")
 
     .def_property_readonly("version",
         &BuildToolVersion::version,
--- a/include/LIEF/MachO/BuildVersion.hpp
+++ b/include/LIEF/MachO/BuildVersion.hpp
@@ -29,7 +29,7 @@ namespace LIEF {
 namespace MachO {
 
 
-class BuildToolVersion : public LIEF::Object {
+class LIEF_API BuildToolVersion : public LIEF::Object {
   public:
   //! @brief Version is an array of **3** integers
   using version_t = std::array<uint32_t, 3>;


romainthomas:
Humm it's already done here: https://github.com/lief-project/LIEF/blob/ab349aadc613fbd04ebc89e1c5e518bcf8334073/api/python/MachO/objects/pyBuildVersion.cpp#L45-L73

mingwandroid:
Thanks. Something else is going on with my builds then. I get the same failure on Linux and macOS. I'll ping with more details later.

mingwandroid:
Ah, I think the problem might be a typo ("BuildeVersion") in https://github.com/lief-project/LIEF/commit/6f967238#diff-48d65da68562e9918a1f502eb47ff2faR53

mingwandroid:
Well, this is probably OK, but it does not fix my problem. I added some tracing for two symbols, one that is OK and also for this problematic one, and in my log I see:

```
x86_64-conda_cos6-linux-gnu/bin/ld: CMakeFiles/LIB_LIEF.dir/src/MachO/Binary.cpp.o: reference to _ZTIN4LIEF5MachO12BuildVersionE
x86_64-conda_cos6-linux-gnu/bin/ld: CMakeFiles/LIB_LIEF.dir/src/MachO/BinaryParser.cpp.o: reference to _ZTIN4LIEF5MachO12BuildVersionE
x86_64-conda_cos6-linux-gnu/bin/ld: CMakeFiles/LIB_LIEF.dir/src/MachO/Builder.cpp.o: reference to _ZTIN4LIEF5MachO12BuildVersionE
x86_64-conda_cos6-linux-gnu/bin/ld: CMakeFiles/LIB_LIEF.dir/src/MachO/BuildVersion.cpp.o: definition of _ZTIN4LIEF5MachO16BuildToolVersionE
x86_64-conda_cos6-linux-gnu/bin/ld: CMakeFiles/LIB_LIEF.dir/src/MachO/BuildVersion.cpp.o: definition of _ZTIN4LIEF5MachO12BuildVersionE
```

So I think the lack of any reference to this causes it to get stripped from the DSO.

romainthomas:
You are right: I also forgot to export the symbol with ``LIEF_API``: https://github.com/lief-project/LIEF/blob/eeddc383938f4262bc1042f98ddeec150a4fca8d/include/LIEF/MachO/BuildVersion.hpp#L32
```cpp
class LIEF_API BuildToolVersion : public LIEF::Object {
```

mingwandroid:
Thank you @romainthomas! I will test with just these two changes and report back to you.

mingwandroid:
Do you want me to update this PR with this change too?

mingwandroid:
Working now. Thank you for the support.

romainthomas:
Thanks for the PR!

PR #382, opened 2020-03-09
Binaries generated with --emit-relocs have dynamic and section relocations

aeflores:
An ELF binary generated with the linker option `--emit-relocs` (https://sourceware.org/binutils/docs/ld/Options.html) will have some dynamic relocations (in `.rela.dyn`) and many more "section" relocations
in e.g. `.rela.text`.

Checking that we do not have any relocation as a condition to parse section relocations makes us miss all those additional relocations. This pull request just removes this check.

### DIFF
--- a/src/ELF/Parser.tcc
+++ b/src/ELF/Parser.tcc
@@ -430,25 +430,24 @@ void Parser::parse_binary(void) {
   }
 
   // Try to parse using sections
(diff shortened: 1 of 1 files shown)


romainthomas:
Hello @aeflores 
Thanks for the PR but I missed the fact that Linux tests are broken because of your changes.
Could you check what is wrong ?

PR #385, opened 2020-03-14
parse section relocations but do not parse dynamic relocations twice.

aeflores:
An ELF binary generated with the linker option --emit-relocs (https://sourceware.org/binutils/docs/ld/Options.html) will have some dynamic relocations (in .rela.dyn) and many more "section" relocations in e.g. .rela.text.

Those additional relocation sections are only there for debugging purposes and are not allocated.
This change always parses non-allocated relocation sections. For allocated sections, it only parses them if at the beginning we do not have any relocations (mimicking the old behavior). This is to take into account cases like `test_parser.py TestCorrupted`.

### DIFF
--- a/src/ELF/Parser.tcc
+++ b/src/ELF/Parser.tcc
@@ -430,23 +430,26 @@ void Parser::parse_binary(void) {
   }
 
   // Try to parse using sections
-  if (this->binary_->relocations_.size() == 0) {
-    for (const Section& section : this->binary_->sections()) {
-
-      try {
-        if (section.type() == ELF_SECTION_TYPES::SHT_REL) {
-
-          this->parse_section_relocations<ELF_T, typename ELF_T::Elf_Rel>(section);
-        }
-        else if (section.type() == ELF_SECTION_TYPES::SHT_RELA) {
-          this->parse_section_relocations<ELF_T, typename ELF_T::Elf_Rela>(section);
-        }
-
-      } catch (const exception& e) {
-        LOG(WARNING) << "Unable to parse relocations from section '"
-                     << section.name() << "'"
-                     << " (" << e.what() << ")";
+  // If we don't have any relocations, we parse all relocation sections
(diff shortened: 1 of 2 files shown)


aeflores:
@romainthomas I would also like to add a test with a binary compiled with `-Wl,--emit-relocs` so the change is actually tested. How should I go about doing that?
Should I make a pull request to https://github.com/lief-project/samples ?

romainthomas:
@alvarofe could you send me by email the source code of your test ?

aeflores:
You can just compile https://github.com/lief-project/samples/blob/master/ELF/hello_c.c as follows:
```
gcc hello_c.c  -Wl,--emit-relocs -o hello-with-relocs
```
You can check that it has additional relocations with `readelf hello-with-relocs -r`

romainthomas:
I uploaded the sample: https://github.com/lief-project/samples/commit/aa3629caaa709494f8f5424e9f24b0c19f378a48

aeflores:
Thanks! I added a test for that binary.

aeflores:
I just realized some of the tests failed! but it seems to be unrelated to my changes. Rather, it seems that the new binary example is not well parsed for some other reason.

The failed test is EXAMPLES_CPP_elf_reader_ELF64_x86-64_hello-with-relocs.bin. I tried this locally and I get an exception thrown at `LIEF::ELF::Binary::eh_frame_functions()`
The error message says

 ```
Try to read 0x1 bytes from 0x2014 (2015) which is bigger than the binary's size
``` 
but 0x2014 is the address (and offset) of the `.eh_frame_hdr`, i.e. it is a legal offset within the binary.

@romainthomas  any idea of what is going on?

romainthomas:
Ok I see why it fails. I'll handle that this week and merge your PR as well (sorry for the delay)

romainthomas:
I have the fix, I'll push it tomorrow.

PR #389, opened 2020-04-05
fix resource parsing bug

kohnakagawa:
In the current implementation of LIEF, `ResourcesManager::dialogs` does not parse the `DLGITEMTEMPLATEEX` structure correctly. This pull request fixes this issue.

https://github.com/lief-project/LIEF/blob/1014ba0ad1a3253e1a08aa0e5f30d3533638013f/src/PE/ResourcesManager.cpp#L1136-L1140

`stream.align(sizeof(uint32_t));` is not nessecary. `windowsClass` could be a null-terminated Unicode string, so the start address of `title` is not always 32 bytes aligned.

This causes the parsing error for some PE files. For example, [NVIDIA Control Panel Application, 8.1.940.0](https://www.virustotal.com/gui/file/ea8b7d9191ea9b6d0d1991f61240285582e3fb1f7d5c08ccf53eb7e3260eae08/detection), the following script

```python
import lief
lief.to_json(lief.parse("nvcplui.exe"))
```

throws `'_pylief.read_out_of_bound'>, read_out_of_bound("Try to read 0x2 bytes from 0x45a (45c) which is bigger than the binary's size",)` exception. This pull request fixes the above parsing error.

It also includes other minor fixes regarding the parsing of `DLGITEMTEMPLATEEX` structure.

Could you merge this pull request?

### DIFF
--- a/src/PE/ResourcesManager.cpp
+++ b/src/PE/ResourcesManager.cpp
@@ -1107,7 +1107,7 @@ std::vector<ResourceDialog> ResourcesManager::dialogs(void) const {
         // =====
         for (size_t i = 0; i < header->nbof_items; ++i) {
           stream.align(sizeof(uint32_t));
-          VLOG(VDEBUG) << "item[" << std::dec << i << "] offset: 0x" << std::hex << stream.pos();
+          VLOG(VDEBUG) << "item[" << std::dec << i << "] offset: " << std::hex << stream.pos();
           ResourceDialogItem dialog_item;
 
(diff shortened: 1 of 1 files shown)


romainthomas:
Thanks @kohnakagawa for this fix! 

Because of a bug detected with the PR #385, CI fail so I'll first fix the issue on the binary ``ELF64_x86-64_hello-with-relocs.bin`` and then I'll merge your PR.

Issue #392, opened 2020-04-09
ELF executable to library demo does not work (can not reproduce documented steps)


suletm:
**Describe the bug**
Demo instructions [here](https://lief.quarkslab.com/doc/latest/tutorials/08_elf_bin2lib.html
) are not reproducible.

vagrant@ubuntu-bionic:/demo$ ./instrument.bin
Error dlsym(): ./libcrackme101.so: undefined symbol: check
vagrant@ubuntu-bionic:/demo$ nm libcrackme101.so |grep check
000000000000371a T check
vagrant@ubuntu-bionic:/demo$ 

**To Reproduce**

To reproduce follow instructions from here: https://lief.quarkslab.com/doc/latest/tutorials/08_elf_bin2lib.html

**Expected behavior**
Demo should work.

**Environment (please complete the following information):**
 - System and Version Distributor ID:	Ubuntu 18.04.2 LTS
 - Target format: ELF
 - LIEF commit version: 0.10.1-bfe5414

**Additional context**
I am using python3 version: 3.6.9. Please note that the demo mentions function named check_found, but the code actually does not have that function. The function we actually export is: check().

romainthomas:
I investigated the issue and it seems to come from the ``GNU_HASH`` table reconstructed by LIEF. We can notice that if we compile the crackme with ``-Wl,--hash-style=sysv`` it works as expected. 

Moreover, the new version of ``glibc``: v2.31 does not allow to ``dlopen`` a binary in which the ``DT_FLAGS_1`` value contains ``PIE``.

> [24930] dynamic-link: dlopen of PIE executable can result in _dl_allocate_tls_init assertion failure

Therefore, on recent Linux system the patch also requires this modification (can be improved):

```python
app[lief.ELF.DYNAMIC_TAGS.FLAGS_1].value = 0
```

suletm:
The first recommendation (--hash-style=sysv) fixed the issue. I did not have to apply the second one since my GLIBC appears to be < 2.31 (based on quick look at readelf -a /path/to/libc) on my system.

Thank you very much for quickly looking at this one. I proposed a patch to fix the documentation:

https://github.com/lief-project/LIEF/compare/master...suletm:patch-1

romainthomas:
See: #397

PR #393, opened 2020-04-11
To fix documentation issue #392

suletm:
1. Inconsistent function name fix.
2. Compilation fix as discussed in #392


romainthomas:
Actually I have to fix a bug in LIEF to support the GNU Hash in this case. Using ``sysv`` hash is just a workaround.

PR #406, opened 2020-05-01
fix machine type enum

kohnakagawa:
Some machine type enums (IMAGE_FILE_MACHINE_AM33 and IMAGE_FILE_MACHINE_RISCV128) are wrong according to https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#machine-types. This pull request fixes this issue.


romainthomas:
Thanks for the fix

PR #432, opened 2020-06-25
fix ResourcesManager to parse string table entry

kohnakagawa:
This pull request fixes ResourcesManager to parse string table entry.

LIEF now can parse string entry contained in PE file as follows.

```python
sample_path = get_sample('PE/PE64_x86-64_binary_WinApp.exe')
mfc = lief.parse(sample_path)
resources_manager = mfc.resources_manager
string_table = resources_manager.string_table
self.assertEqual(string_table[0].name, "WinApp")
self.assertEqual(string_table[0].length, 6)
```


romainthomas:
Thank you!

PR #435, opened 2020-07-03
Fixed section extension shifting

tsarpaul:
section extend is calculated wrong and can result in relocations/symbols being shifted when they shouldn't be. The from_address is where the extension begins, and should be the end of the section.

You can see extend_segment is calculated properly:
https://github.com/lief-project/LIEF/blob/master/src/ELF/Binary.tcc#L572


tsarpaul:
Did not mean to add the 2nd commit. I will fix this soon.

romainthomas:
I guess it is fixed now but if you still face to this issue feel free to re-open

PR #450, opened 2020-08-19
Fix build issue for python-32 on x64 windows

tacesrever:
env: python3.8-win32 on x64 win10, vs 2019
when running `setup.py install`, the default compiler will be x64.
and seems #408 forget to add undef in PE/undef.h, which cause issue when building from master.


api/python/pyLIEF.cpp
```diff
@@ -50,7 +50,9 @@
 #include "platforms/android/pyAndroid.hpp"
 
 
+py::gil_scoped_acquire *_init_acquire = new py::gil_scoped_acquire();
```
romainthomas, line 53: Could you explain the purpose of these modifications ?

romainthomas: My bad, I didn't read the commit message

PR #467, opened 2020-09-20
Improve PE authenticode parsing logic

kohnakagawa:
@romainthomas
I improved PE authenticode parsing logic (sorry for so big patch). This pull-request includes:

- implementation of UnauthenticatedAttributes parser (nested signature and countersignature can be parsed)
- implementation of SpcIndirectDataContent parser
- refactoring of existing code

@recvfrom 
Hi. I heard that you use LIEF for Authenticde parsing. Could you test this?


kohnakagawa:
Sorry, some CI tests failed. I will fix this issue soon.

recvfrom:
@kohnakagawa nice work on this! yep I'll test this and let you know how it goes

romainthomas:
Thanks @kohnakagawa for the nice PR :+1: (as always). I'll take a closer look this week!

romainthomas:
@kohnakagawa I commented your PR (I can handle the modification if you prefer)

src/PE/json.cpp
```diff
@@ -788,27 +797,84 @@ void JsonVisitor::visit(const x509& x509) {
...
 void JsonVisitor::visit(const ContentInfo& contentinfo) {
-  this->node_["content_type"]     = contentinfo.content_type();
-  this->node_["type"]             = contentinfo.type();
-  this->node_["digest_algorithm"] = contentinfo.digest_algorithm();
+  this->node_["content_type"] = contentinfo.content_type();
+}
+
+void JsonVisitor::visit(const SpcIndirectDataContent& spc_indirect_data_content) {
+  this->node_["content_type"] = spc_indirect_data_content.content_type();
+  this->node_["digset_algorithm"] = spc_indirect_data_content.digest_algorithm();
```
recvfrom, line 839: this should be `digest_algorithm`

api/python/PE/objects/signature/pyContentInfo.cpp
```diff
@@ -40,18 +41,6 @@ void create<ContentInfo>(py::module& m) {
...
-    .def_property_readonly("digest_algorithm",
-        &ContentInfo::digest_algorithm,
-        "Algorithm (OID) used to hash the file. This value should match SignerInfo.digest_algorithm and Signature.digest_algorithm")
-
-
-    .def_property_readonly("digest",
-        &ContentInfo::digest,
-        "The digest")
-
     .def_property_readonly("raw",
```
recvfrom, line 44: Can we move this `raw` field into `SpcIndirectDataContent` like the three above?  It probably makes more sense to live there anyway, since that field currently gets populated with only the SpcIndirectDataContent bytes

https://github.com/lief-project/LIEF/blob/657d589fae9949ad48b6e074517299853b4a99ba/src/PE/signature/SignatureParser.cpp#L173-L192

With the ContentInfo no longer being passed back via `Signature.content_info` (now it's a `SpcIndirectDataContent` that usually gets passed back) right now it doesn't seem like it's possible to access this `raw` field if you have a `Signature` object...

Access to the raw bytes is needed when validating the Authenticode signature... I should probably add Authenticode validation logic as a LIEF test case so it's more clear how these raw fields can be used. 🤔

recvfrom:
I tested my scripts (python) against a bunch of samples and it worked for almost all of them, but python seg faults when scanning https://www.virustotal.com/gui/file/0ee196bb23f0eafe4f61d30bf6c676fd7365cb12ae66a6bde278851e91901ac1/details...

The following is enough to trigger the issue:
```
fd = open('0ee196bb23f0eafe4f61d30bf6c676fd7365cb12ae66a6bde278851e91901ac1', 'rb')
data = fd.read()
fd.close()
_ = lief.parse(list(data))
```

I don't get any memory issues when running the executable with HelloLIEF under valgrind... any thoughts as to what might be causing this?

kohnakagawa:
@romainthomas  @recvfrom 

Sorry for the slow response. Recently I have been busy at work. I'll check your comments and revise my pull-request.

romainthomas:
I refactored the whole logic in the branch [enhancement/pe-authenticode](https://github.com/lief-project/LIEF/tree/enhancement/pe-authenticode)
and the API looks like this now:

https://github.com/lief-project/LIEF/blob/enhancement/pe-authenticode/examples/python/pe_sign/authenticode.py

Especially, there is now:

https://github.com/lief-project/LIEF/blob/b06a99d48dd9e5df0674553eea47862ab4f27e8c/src/PE/signature/Signature.cpp#L156-L263

and 

https://github.com/lief-project/LIEF/blob/b06a99d48dd9e5df0674553eea47862ab4f27e8c/src/PE/Binary.cpp#L1205-L1242

kohnakagawa:
@romainthomas Sorry for the late reply. I’m sorry, but I could not handle the problem of this pull request (https://github.com/lief-project/LIEF/pull/467) to prepare other projects. Thank you for your reimplementation of PE authenticode parsing logic. I read your implementation, and this implementation is better than mine!

romainthomas:
No worry @kohnakagawa :)

PR #475, opened 2020-10-11
Add simple cbegin and cend implementation.

Clcanny:


Clcanny:
@kohnakagawa Hello, please forget my poor English. Please take some time to review this change.

After changing, LIEF can be compiled at the following environment:

```bash
# uname -r
4.9.0-9-amd64
# cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 9 (stretch)"
# cmake --version
cmake version 3.7.2
# g++ --version
g++ (Debian 6.3.0-18+deb9u1) 6.3.0 20170516
```

romainthomas:
Hello @Clcanny 
I don't understand the purpose of this include. Why would you want to include this C++14 feature in LIEF ?

Clcanny:
> Hello @Clcanny
> I don't understand the purpose of this include. Why would you want to include this C++14 feature in LIEF ?

Hi, actually I want to use C++14. But it seems that C++ 14 isn't default option in LIEF.

romainthomas:
Indeed, but I think that it's up to you to include this file in you project instead of forcing LIEF to use it (since LIEF does not need it)

Clcanny:
Okay, thanks a lot for your reply.

PR #487, opened 2020-11-10
add mips to the elf architecture mapping.

aeflores:
Map the architecture in the elf header for mips with the abstract architecture.


romainthomas:
Thanks!

PR #493, opened 2020-11-22
Sort static symbols in Builder for ELF.

Clcanny:
```cpp
// main.cpp
namespace {
int foo = 1;
}  // anonymous namespace

int bar = 2;

int main() {
}
```

```cpp
// test.cpp
#include <LIEF/ELF.hpp>
#include <memory>
#include <string>

int main(int argc, char* argv[]) {
  bool smart_mode = argc == 2 && argv[1] == std::string("true");
  std::unique_ptr<LIEF::ELF::Binary> binary(LIEF::ELF::Parser::parse("main"));
  std::string main_name = "main.cpp";
  std::string foo_name = "_ZN12_GLOBAL__N_13fooE";
  LIEF::ELF::Symbol main_sym = binary->get_static_symbol(main_name);
  LIEF::ELF::Symbol foo_sym = binary->get_static_symbol(foo_name);
  binary->remove_static_symbol(main_name);
  binary->remove_static_symbol(foo_name);
  binary->add_static_symbol(main_sym, smart_mode);
  binary->add_static_symbol(foo_sym, smart_mode);
  binary->write("main-modified");
}
```

```makefile
all :
        # Clean.
        rm -f main test main-modified
        # Compile.
        gcc main.cpp -O0 -ggdb -o main
        g++ -std=c++11 -O0 -ggdb test.cpp                                                    \
                -I/usr/include/LIEF-0.11.0-Linux/include -L/usr/lib/LIEF-0.11.0-Linux/lib -lLIEF \
        -o test
```

If run without smart mode, readelf will complain:

```bash
# ./test
# readelf --symbols main-modified | grep "Warning"
readelf: Warning: local symbol 71 found at index >= .symtab's sh_info value of 52
readelf: Warning: local symbol 72 found at index >= .symtab's sh_info value of 52
```

The reasons why readelf complains are:

1. Non-exported symbols should be sorted before exported symbols.
2. Information of .symtab section header records the first index of exported symbol.

If run with smart mode, readelf won't complain:

```bash
# ./test true
# readelf --symbols main-modified | grep "Warning"
```

1. I don't test it in Python because I don't know how to install it to Python.
2. It isn't smart enough, but I think it is a begining.

### DIFF
--- a/src/ELF/Builder.tcc
+++ b/src/ELF/Builder.tcc
@@ -491,6 +491,30 @@ void Builder::build_static_symbols(void) {
   //clear
   //symbol_section.content(std::vector<uint8_t>(symbol_section.content().size(), 0));
 
+  auto it_begin = std::begin(this->binary_->static_symbols_);
+  auto it_end = std::end(this->binary_->static_symbols_);
+  std::stable_sort(it_begin, it_end, [](const Symbol* lhs, const Symbol* rhs) {
+    if (lhs->binding() == SYMBOL_BINDINGS::STB_LOCAL &&
+        (rhs->binding() == SYMBOL_BINDINGS::STB_GLOBAL ||
+         rhs->binding() == SYMBOL_BINDINGS::STB_WEAK)) {
+      return true;
+    }
+    return false;
+  });
+  auto it_first_exported_symbol =
+      std::find_if(it_begin, it_end, [](const Symbol* sym) {
+        return sym->binding() == SYMBOL_BINDINGS::STB_GLOBAL ||
+               sym->binding() == SYMBOL_BINDINGS::STB_WEAK;
+      });
+  uint32_t first_exported_symbol_index =
+      static_cast<uint32_t>(std::distance(it_begin, it_first_exported_symbol));
+  if (first_exported_symbol_index != symbol_section.information()) {
+    LIEF_WARN("information of .symtab section changes from {:d} to {:d}",
+              symbol_section.information(),
+              first_exported_symbol_index);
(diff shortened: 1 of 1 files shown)


Clcanny:
@romainthomas Hello, please review this change. Thanks a lot for your reviewing! (And please forget my poor English.)

romainthomas:
Hello @Clcanny 
Thanks for this PR, I think that it would be better to make this kind of sort in the builder as it already done for the dynamic symbols:

https://github.com/lief-project/LIEF/blob/2145d79efb361eff87a3e5b05d9bf84720c491f6/src/ELF/Builder.tcc#L916-L923

Clcanny:
> I think that it would be better to make this kind of sort in the builder as it already done for the dynamic symbols:
> 
> https://github.com/lief-project/LIEF/blob/2145d79efb361eff87a3e5b05d9bf84720c491f6/src/ELF/Builder.tcc#L916-L923

Hello @romainthomas , thanks for your review. I think the same logic should work for static symbols too.

Should I add the same logic for static symbols in the builder?

romainthomas:
> Should I add the same logic for static symbols in the builder?

Yes, in the ``build_static_symbol()`` the first processing should be sorting the symbol

Clcanny:
> Yes, in the `build_static_symbol()` the first processing should be sorting the symbol

Hello, I have changed the commit.

romainthomas:
Merged with https://github.com/lief-project/LIEF/commit/588a7a3a04bfce43c88714954bbc27f1bca8d0d7

Clcanny:
Thank you!

PR #497, opened 2020-11-24
Don't do relocation and etc. when extend not loaded section.

Clcanny:
```cpp
// main.cpp
int main() {}
```

```cpp
// extend.cpp
#include <LIEF/ELF.hpp>
#include <memory>

int main() {
  std::unique_ptr<LIEF::ELF::Binary> binary(LIEF::ELF::Parser::parse("main"));
  const LIEF::ELF::Section& section = binary->get_section(".strtab");
  binary->extend(section, 64);
  std::cout << binary->get_section(".strtab").size() << std::endl;
  binary->write("main-modified");
}
```

```makefile
all :
        rm -f main extend main-modified
        gcc main.cpp -O0 -ggdb -o main
        g++ -std=c++11 -O0 -ggdb extend.cpp                                                                       \
                -I/usr/include/LIEF-0.11.0-Linux/include -L/usr/lib/LIEF-0.11.0-Linux/lib -lLIEF \
        -o extend
        ./extend
        chmod u+x main-modified
        ./main-modified
```

The above code will complain:

```bash
NEEDED not supported
INIT_ARRAYSZ not supported
FINI_ARRAYSZ not supported
STRSZ not supported
SYMENT not supported
DEBUG not supported
RELASZ not supported
RELAENT not supported
FLAGS_1 not supported
VERNEEDNUM not supported
RELACOUNT not supported
NULL not supported
NULL not supported
NULL not supported
NULL not supported
NULL not supported
terminate called after throwing an instance of 'LIEF::not_found'
  what():  Unable to find the segment associated with the address: 0x201060
Makefile:2: recipe for target 'all' failed
make: *** [all] Aborted
```

But, sections that won't be loaded into memory aren't needed to do some operations, such as relocations.

### DIFF
--- a/src/ELF/Binary.cpp
+++ b/src/ELF/Binary.cpp
@@ -1344,6 +1344,7 @@ Section& Binary::extend(const Section& section, uint64_t size) {
 
   uint64_t from_offset  = section_to_extend->offset() + section_to_extend->size();
   uint64_t from_address = section_to_extend->virtual_address() + section_to_extend->size();
+  bool section_loaded   = section_to_extend->virtual_address() != 0;
   uint64_t shift        = size;
 
   this->datahandler_->make_hole(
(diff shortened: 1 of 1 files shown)


Clcanny:
@romainthomas Hello, please review this change! Thanks a lot!

romainthomas:
Looks good to me. Could you squash your commits ?

Clcanny:
> Looks good to me. Could you squash your commits ?

Yes, I can. Thank you!

romainthomas:
You can take a look at: 

https://github.com/lief-project/LIEF/blob/2145d79efb361eff87a3e5b05d9bf84720c491f6/tests/elf/add_section.py#L31-L55

Clcanny:
I have squashed the commits. And I am not clear should we call `shift_dynamic_entries` when `section_loaded` is false.

PR #498, opened 2020-11-26
Simplify find_package mechanics

### LINKED ISSUE #375: CMake integration fails to find LIEF_LIBRARY
**Describe the bug**
When built and installed on Linux with:
```
cd build/
cmake -DCMAKE_BUILD_TYPE=Release ..
make
cd api/python
python setup.py build
make DESTDIR=/usr install
python setup.py install --root="${pkgdir}" --optimize=1
```
The `/usr/share/LIEF/cmake/FindLIEF.cmake` file is unable to locate `LIEF_LIBRARY` (which is located in `/usr/lib/libLIEF.so`.  This makes it impossible to use `find_package(LIEF)` in another CMake script.  Add the following to FindLIEF.cmake works around the problem but is not a solution.

```
@@ -108,6 +108,9 @@ if(NOT LIEF_LIBRARY)
   set(LIEF_LIBRARY ${_LIEF_LIBRARY})
 endif()
 
+if(NOT LIEF_LIBRARY)
+  set(LIEF_LIBRARY "/usr/lib/libLIEF.so")
+endif()
 
 unset(LIEF_NAMES)
 ```

**To Reproduce**
Install as above and try to use with find_package.  This can also be done easily on Arch Linux using the aur package at https://aur.archlinux.org/packages/lief.

**Expected behavior**
I would hope for `find_package(LIEF)` in CMake to work.

 - System and Version : Arch Linux
 - Target format NA
 - LIEF commit version: 0.9.0

**Additional context**
Nothing else.

Thanks!

kwarrick:
Generate a `LIEFConfig.cmake` and `LIEFConfigVersion.cmake` to simplify CMake integration.

Integration will now only require a single `find_package` and you will no longer have to specify `-DLIEF_ROOT`.

Fixes #375.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,7 @@ endif()
 include(CheckCXXCompilerFlag)
 include(CheckCCompilerFlag)
 include(ExternalProject)
+include(CMakePackageConfigHelpers)
 
 if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/.git"
    AND IS_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/.git")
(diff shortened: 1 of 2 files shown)


romainthomas:
Thanks!

PR #519, opened 2021-01-06
Calculate MachO binary available_command_space_

achow101:
This value was always set to 0, so space would always be allocated in
the binary for more commands. However it is not necessary to do this as
there will often be space available. This commit adds a simple way to
calculate that during parsing.

This method is taken from Apple's cctools package, specifically the
codesign_allocate tool. See https://github.com/tpoechtrager/cctools-port/blob/466063c7f7486762a5a8f0d98c48dc5ffbe42f74/cctools/misc/codesign_allocate.c#L861


romainthomas:
Thanks for the fix. It has been merged with: https://github.com/lief-project/LIEF/commit/8f46263c17671022beb5be448f5aebd130ccb86e

PR #520, opened 2021-01-06
MachO: Don't set strings table size based on linkedit size

achow101:
Given that the strings table size cannot grow, and a smaller strings
table is padded out to the original size, set the strings table size to
symbol_command->strings_size().

Additionally, the previous code would cause the strings table to be
expanded to the rest of the linkedit segment which breaks anything that
may add data to a linkedit segment, or any data that already exists
after the strings table in the linkedit segment. Notably this issue can
be seen when working with MachO binaries that have code signatures as
these are placed at the end of the linkedit segment.


romainthomas:
Fixed since then

PR #522, opened 2021-01-15
Fix the number of entries of MIPS_RELOCS

kohnakagawa:
Hi. @romainthomas 

This is a quite small fix.

The number of entries of RELOC_MIPS seems to be wrong. Due to this, master branch HEAD cannot be compiled on my environment. This pull request fixes this issue.


romainthomas:
You are completely right and I don't know why CI didn't catch it ...

kohnakagawa:
I think so too...

I think we need to investigate why this compilation error was not occurring in CI.

PR #528, opened 2021-01-19
Fix IAT Reconstruction

### LINKED ISSUE #523: Reconstructing IAT many times can break PE files
**Describe the bug**
Applying IAT reconstruction by `patch_imports(true)` and `build_imports(true)` many times can produce broken PE files.

**To Reproduce**
1. Make a PE file by compiling the following code.
```cpp
#include <stdio.h>

int main() {
	printf("Hello, World!");
	return 0;
}
```
(I used Microsoft Visual Studio Professional 2019 Version 16.3.9 to compile it.)
Assume that this produced "original.exe".

2. Clone the LIEF repository and modify LIEF/examples/cpp/pe_builder.cpp to the following code.
```cpp
#include <iostream>
#include <memory>
#include <fstream>
#include <algorithm>
#include <iterator>

#include <LIEF/PE.hpp>

using namespace LIEF::PE;
int main(int argc, char **argv) {
  std::cout << "PE Rebuilder" << std::endl;
  if (argc != 3) {
    std::cerr << "Usage: " << argv[0] << " <Input Binary> <Output Binary>" << std::endl;
    return -1;
  }

  std::unique_ptr<Binary> binary{Parser::parse(argv[1])};
  binary->optional_header().remove(DLL_CHARACTERISTICS::IMAGE_DLL_CHARACTERISTICS_DYNAMIC_BASE);
  Builder builder{binary.get()};

  builder
    .build_imports(true)
    .patch_imports(true)
    .build_tls(false)
    .build_resources(false);

  builder.build();
  builder.write(argv[2]);
  std::cout << binary->name() << std::endl;

  return 0;
}
```

3. Build LIEF and the example codes including the above modified LIEF/examples/cpp/pe_builder.cpp. This makes modified version of pe_builder.

4. Applyfing IAT reconstruction by the modified pe_builder 9 times. For example,
```
./pe_builder original.exe 2.exe
./pe_builder 2.exe 3.exe
./pe_builder 3.exe 4.exe
...
./pe_builder 9.exe 10.exe
```

5. Run the 10.exe and you will see that it exits abnormally.

**Expected behavior**
You will see "Hello, World!".

 - System and Version :  Ubuntu 18.04 (Building LIEF and modifying original.exe) and Ubuntu 18.04 on WSL1 on Windows 10 1909 (running the 10.exe)
 - Target format (PE, ELF, Mach-O) : PE
 - LIEF commit version: 0.11.0-cd286e1

**Additional context**
This behavior seems to be due to the current implementation of finding the "original" import section logic.
Look at the LIEF/src/PE/Builder.tcc.
```cpp
  // As add_section will change DATA_DIRECTORY::IMPORT_TABLE we have to save it before
  uint32_t offset_imports  = this->binary_->rva_to_offset(this->binary_->data_directory(DATA_DIRECTORY::IMPORT_TABLE).RVA());
  Section& import_section = this->binary_->add_section(new_import_section, PE_SECTION_TYPES::IMPORT);


  // Patch the original IAT with the address of the associated trampoline
  if (this->patch_imports_) {
    Section& original_import = this->binary_->section_from_offset(offset_imports);
    std::vector<uint8_t> import_content  = original_import.content();
    uint32_t roffset_import = offset_imports - original_import.offset();
```
In some cases, `original_import` is _not_ corresponding to the "original" import section.
The IAT reconstruction makes an additional section so that an additional section header will be added, too.
This increase the header size and can move sections backward. In this case, `original_import` returns the previous section of the "original" import section.

I am creating a PR but some tests failed at commit d1ad5ab1de4fe16d9fee00c55ebb8d1aeb7ee519 on Ubuntu 18.04 on EC2 without any modification, so I am investigating the issue.

Catminusminus:
This will be fix #523

Currently, `this->binary_->section_from_offset(offset_imports)` is used as the original import section. 
However, if the "backward shift" (explained in #523 ) happens, this returns the previous section of the original import section. So I try another way to obtain the original import section.

I introduce `PE_SECTION_TYPES::OLD_IMPORT` type to mark the original import section.
The type will be added to the target section when the `PE_SECTION_TYPES::IMPORT` type will be removed from it.

After adding the new import section, we can find the original import section by  looking to see if the type is `PE_SECTION_TYPES::OLD_IMPORT`. 

If we can not find the original import section, it will fall back to the current implementation, that is, `this->binary_->section_from_offset(offset_imports)` will be used.

macOS and iOS CI jobs currently failed, but I suspect that they are due to the GitHub Actions' issue (https://github.com/actions/virtual-environments/issues/841).
In fact, I see the "(X) This check failed" message and these jobs are unstable...
https://github.com/Catminusminus/LIEF/actions/runs/492792964

Sorry if I my suspicions are wrong. I'll re-run the jobs to find out what's wrong, anyway.

### DIFF
--- a/include/LIEF/PE/enums.hpp.in
+++ b/include/LIEF/PE/enums.hpp.in
@@ -23,7 +23,8 @@ enum class PE_SECTION_TYPES : uint8_t {
   EXPORT     = 7,
   DEBUG      = 8,
   LOAD_CONFIG = 9,
-  UNKNOWN     = 10
+  UNKNOWN     = 10,
+  OLD_IMPORT = 100
 };
 
 enum class PE_TYPE : uint16_t {
--- a/src/PE/Builder.tcc
+++ b/src/PE/Builder.tcc
@@ -168,20 +168,28 @@ void Builder::build_import_table(void) {
   // Remove 'import' type from the original section
   if (it_import_section != std::end(this->binary_->sections_)) {
     (*it_import_section)->remove_type(PE_SECTION_TYPES::IMPORT);
+    (*it_import_section)->add_type(PE_SECTION_TYPES::OLD_IMPORT);
   }
 
   // As add_section will change DATA_DIRECTORY::IMPORT_TABLE we have to save it before
   uint32_t offset_imports  = this->binary_->rva_to_offset(this->binary_->data_directory(DATA_DIRECTORY::IMPORT_TABLE).RVA());
+  auto offset_imports_in_section = offset_imports - this->binary_->section_from_offset(offset_imports).offset();
   Section& import_section = this->binary_->add_section(new_import_section, PE_SECTION_TYPES::IMPORT);
-
 
   // Patch the original IAT with the address of the associated trampoline
   if (this->patch_imports_) {
-    Section& original_import = this->binary_->section_from_offset(offset_imports);
+    auto&& original_import_section = std::find_if(
+        std::begin(this->binary_->sections_),
+        std::end(this->binary_->sections_),
+        [] (const Section* section) {
+          return section != nullptr and section->is_type(PE_SECTION_TYPES::OLD_IMPORT);
+        });
+    (*original_import_section)->remove_type(PE_SECTION_TYPES::OLD_IMPORT);
+    auto original_import_offset = (*original_import_section)->pointerto_raw_data();
+    Section& original_import = this->binary_->section_from_offset(original_import_offset);
     std::vector<uint8_t> import_content  = original_import.content();
-    uint32_t roffset_import = offset_imports - original_import.offset();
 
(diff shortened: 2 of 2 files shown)


Catminusminus:
I found this PR works for C++, but doesn't work for Python API. That is, applying the following code 9 times creates an invalid PE file.
```python
    binary = lief.PE.parse(input_path)
    binary.optional_header.dll_characteristics &= (
        ~lief.PE.DLL_CHARACTERISTICS.DYNAMIC_BASE
    )
    builder = lief.PE.Builder(binary)
    builder.build_imports(True).patch_imports(True)
    builder.build()
    builder.write(output_path)
```
I'm investigating this issue.

Catminusminus:
The above my comment is wrong and I conformed that this PR do work for Python API.
I'm sorry for my misunderstanding.

Issue #517, opened 2021-01-05
add_entry function can create invalid PE files


Catminusminus:
**Describe the bug**
Adding an entry to the import table of PE by `add_entry` function can create an invalid PE file.

**To Reproduce**
1. Make a PE file by compiling the following code.
```
#include <stdio.h>

int main() {
	printf("Hello, World!");
	return 0;
}
```
(I used Microsoft Visual Studio Professional 2019 Version 16.3.9 to compile it.)

2. Make sure VCRUNTIME140.dll is _not_ at the bottom of the imports of the PE file. For example, KERNEL32.dll is the bottom of  the imports in the picture.

![image](https://user-images.githubusercontent.com/37803616/103608795-83673a80-4f5f-11eb-8d85-32e8cca83db3.png)

3.  Adding `wcsrchr` entry to the PE file.
```
import lief

input_path = "sample.exe" # the PE file's path 
output_path = "modified.exe"
binary = lief.PE.parse(input_path)
binary.optional_header.dll_characteristics &= (
    ~lief.PE.DLL_CHARACTERISTICS.DYNAMIC_BASE
)

lib = binary.get_import("VCRUNTIME140.dll")

lib.add_entry("wcsrchr")
builder = lief.PE.Builder(binary)
builder.build_imports(True).patch_imports(True)
builder.build()
builder.write(output_path)
```

4. Run `modified.exe` and you will see that  it exits abnormally.

**Expected behavior**
You will see "Hello, World!".

**Environment (please complete the following information):**
 - System and Version : Ubuntu 18.04 on WSL1 on Windows 10 1909
 - Target format : PE
 - LIEF commit version: 0.11.0-8b1d101

**Additional context**
I found that `__initterm_e` function jumps to `exit` function in `modified.exe` by debugging it. So I suspect that adding an entry to the IAT makes the correspondence between the import function and the trampoline is unintentionally shifted back one by one. 
Actually, if you add `wcsrchr` entry and `strstr` entry to the original PE file, `__initterm_e` function jumps to not `exit` function but `_exit` function.

PR #530, opened 2021-01-20
Fix build_imports

kohnakagawa:
Hi

This pull request fixes the issue (#517).

The current implementation of `build_import_table` does not check the correspondence between the imported functions and the trampolines. So, if we add an import function entry by `add_entry` and build IAT by calling `build_import_table`, its correspondence is shifted back one by one.

I illustrate this situation.

Before adding a new import entry, the correspondence between IAT and trampolines is correct.

図1

However, after adding `wcsrchr` (illustrated by orange color), its correspondence is incorrect. In this case, when we try to call `puts` function, `_cexit` function is called instead.

図2

To fix this issue, I added a new member `manually_added_` to `ImportEntry` class to distinguish whether the entry is manually added or not. If the entry is a manually-added one, linking between the original IAT entry and the trampoline is skipped.

### DIFF
--- a/include/LIEF/PE/ImportEntry.hpp
+++ b/include/LIEF/PE/ImportEntry.hpp
@@ -66,9 +66,11 @@ class LIEF_API ImportEntry : public Object {
   //! @brief **Original** address of the entry in the Import Address Table
   uint64_t iat_address(void) const;
 
+  bool manually_added(void) const;
 
   void name(const std::string& name);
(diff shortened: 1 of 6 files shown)


romainthomas:
Note: there is this working branch: [peimports](https://github.com/lief-project/LIEF/tree/peimports) that tries to improve the way we patch import

PR #543, opened 2021-02-04
Use switch/case statements instead of look-up tables for ordinals_lookup_tables

j-jorge:
The creation of the multiple `static const std::map<uint32_t, const char*>` for lookup tables were popping in my profiler. I believe they could be advantageously replaced with `switch`/`case` statements, preventing allocations at startup and reducing the amount of work for lookups.

Please review this MR for these modifications.


romainthomas:
Nice! I'm completely ok with these changes and I think that now the function is also ``constexpr``.

PR #545, opened 2021-02-11
Fix parse_dynamic_entries to stop iterating at DT_NULL

junghee:
Previously, parse_dynamic_entries iterates the entire dynamic segment.
Ideally, it needs to only iterate dynamic section for dynamic entries.
This fix uses DT_NULL, which marks end of dynamic section.

### DIFF
--- a/src/ELF/Parser.tcc
+++ b/src/ELF/Parser.tcc
@@ -1112,6 +1112,7 @@ void Parser::parse_dynamic_entries(uint64_t offset, uint64_t size) {
 
   Elf_Off dynamic_string_offset = this->get_dynamic_string_table();
 
+  bool end_of_dynamic_section = false;
   this->stream_->setpos(offset);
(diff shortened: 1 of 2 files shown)


romainthomas:
I agree on the fact that we should break on ``DT_NULL`` but I disagree with 
> Ideally, it needs to only iterate dynamic section

Since sections can be stripped/corrupted without altering the execution. 
Would you mind to change the variable ``end_of_dynamic_section`` into ``end_of_dynamic`` ?

junghee:
> 
> 
> I agree on the fact that we should break on `DT_NULL` but I disagree with
> 
> > Ideally, it needs to only iterate dynamic section
> 
> Since sections can be stripped/corrupted without altering the execution.
> Would you mind to change the variable `end_of_dynamic_section` into `end_of_dynamic` ?

Thank you for a quick response. I just pushed a fix.

PR #553, opened 2021-03-06
Set spdlog logging level to spdlog::level::debug when user compile with compile with option '-LIEF_LOGGING_DEBUG=ON'

Clcanny:


romainthomas:
Why ? I mean the user can still use:

https://github.com/lief-project/LIEF/blob/9e084e64ff4de9a68d21190ccb24c1d7762742ef/include/LIEF/logging.hpp#L49

Clcanny:
Yes, the user can still use it like that way. But I think it is counter-intuitive for me that I can't see debug log after setting the flag. So I would like to change it.

romainthomas:
Well, I do not agree but if other users would like to have this behavior, please let a comment.

Issue #418, opened 2020-05-21
Last entry in .data section content is not updated to new offset when segment is added


pdreiter:
**Describe the bug**
Content of .data section, when last 64 bytes is an .rodata address, is not updated when a new segment is added.

**To Reproduce**
I do not have a simple input binary that demonstrates this issue, but : 

1. parse AIS-Lite from https://github.com/trailofbits/cb-multios, compiled with gcc (CC=gcc CXX=g++ build.sh)
2. add new segment to ./AIS-Lite and generate output binary ./added_seg.bin => contents of ./added_seg.bin have mostly been shifted by 0x1000
3. crudely evaluating the last 8 bytes of contents from each the .data section:
```
>>>x=lief.parse("./AIS-Lite")
>>>y=lief.parse("./added_seg.bin")
>>>lx=len(x.get_section(".data").content)
>>>ly=len(y.get_section(".data").content)
>>>print(x.get_section(".data").content[lx-8:])
[43, 65, 0, 0, 52, 65, 0, 0]
>>>print(z.get_section(".data").content[lz-8:])
[43, 81, 0, 0, 52, 65, 0, 0]
```
These addresses correspond to the .rodata offsets `0x41b2` and `0x4134` in the original binary.
In the ./added_seg.bin, the corresponding .rodata offsets for these symbols are `0x51b2` and `0x5134`, but in the actual .data contents of ./added_seg.bin are `0x51b2` and `0x4134`, respectively. Symbols are fine, just the last .data content has not been updated to the new offset.
I'm not sure if this information is relevant, but the problematic global symbol is a const array of char*.  

**Expected behavior**
I'm expecting that if some .data content is updated to the new offset, then all content is updated.
That means this:
```
>>>print(z.get_section(".data").content[lz-8:])
[43, 81, 0, 0, 52, 65, 0, 0]
```
should be:
```
>>>print(z.get_section(".data").content[lz-8:])
[43, 81, 0, 0, 52, **81**, 0, 0]
```

**Environment (please complete the following information):**
 - Ubuntu 19.04
 - Target format: ELF
 - LIEF commit version: 0.10.1-bfe5415

**Additional context**

romainthomas:
I think I understand the issue but the modification seems not generic ? Also, did you check the relocations ?

pdreiter:
I did check the relocation sections, but I will check again.
Anything in particular you want me to report? 
 I will also update with more info about the problematic global symbol.

[...]

PR #568, opened 2021-04-01
resolution for LIEF issue #418

pdreiter:
addresses boundary scenario for relocations whose
relative offsets abut segment_size

Apologies for my miscommunication re: _relocations_ with @romainthomas -- this waylaid the resolution which @romainthomas had speculated.

While 64b compilation did not express same run-time failure as 32b, both 32b and 64b relocation content `value`s do not reflect the `shift`s prior to this change.

Tested on scenario described in this comment and successfully addresses 32b runtime issue - and benign to 64b.
https://github.com/lief-project/LIEF/issues/418#issuecomment-657779791

### DIFF
--- a/src/ELF/Binary.tcc
+++ b/src/ELF/Binary.tcc
@@ -242,7 +242,7 @@ void Binary::patch_addend(Relocation& relocation, uint64_t from, uint64_t shift)
     return;
   }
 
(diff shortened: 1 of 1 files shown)


romainthomas:
Thanks @pdreiter !

PR #570, opened 2021-04-13
Fix sizeof_headers computing

### LINKED ISSUE #569: Rebuild a PE file can make an invalid PE file
**Describe the bug**
Rebuild a PE file using LIEF can produce an invalid PE file that if you run it and you get the message "This application could not be started."

**To Reproduce**
First, build LIEF and the example program `pe_builder`.

For example,
```
git clone https://github.com/lief-project/LIEF.git
cd LIEF
mkdir build
cd build
cmake -DLIEF_PYTHON_API=off -DCMAKE_BUILD_TYPE=Release ..
cmake --build . --target LIB_LIEF --config Release
```
from (https://lief.quarkslab.com//doc/latest/compilation.html)

Then you run `make` in the `LIEF/build/example/cpp` directory, and you will get `pe_builder`.

Next, run
```
./pe_builder 7152c6554215cabb53c75d1cd7d988c10351958fb7d190625e6875e3ce57756e out.exe
```
where the file `7152c6554215cabb53c75d1cd7d988c10351958fb7d190625e6875e3ce57756e` has 7152c6554215cabb53c75d1cd7d988c10351958fb7d190625e6875e3ce57756e as its sha256 hash value.
(Because this is a malware, I cannot provide you this file due to the laws of our country. Sorry.)

Then run `out.exe` in Windows and you will get the message "This application could not be started."

**Expected behavior**
You can start `out.exe`. 

 - System and Version : Ubuntu 20.04
 - Target format PE
 - LIEF commit version: 0.12.0-3b7b5a7

**Additional context**
This caused by the current implementation of `Binary::sizeof_headers`(https://github.com/lief-project/LIEF/blob/master/src/PE/Binary.cpp#L542).
In this method, I think
```
  size += sizeof(pe_data_directory) * (this->data_directories_.size() + 1);
  size += sizeof(pe_section) * (this->sections_.size() + 1);
```
must be
```
  size += sizeof(pe_data_directory) * (this->data_directories_.size());
  size += sizeof(pe_section) * (this->sections_.size());
```

Indeed, the file `7152c6554215cabb53c75d1cd7d988c10351958fb7d190625e6875e3ce57756e` has its header whose size is 1f0 and its sections which starts from 200 (raw address), but `Binary::sizeof_headers` returns 400.

I will submit a PR to fix this.

Catminusminus:
This PR will be fix #569.
Currently, in `Binary::sizeof_headers` method (https://github.com/lief-project/LIEF/blob/master/src/PE/Binary.cpp#L542-L557), `this->data_directories_.size()` and `this->sections_.size()` are added 1.
```
  size += sizeof(pe_data_directory) * (this->data_directories_.size() + 1);
  size += sizeof(pe_section) * (this->sections_.size() + 1);
```
I think these "+1" are unnecessary.

Indeed, the file 7152c6554215cabb53c75d1cd7d988c10351958fb7d190625e6875e3ce57756e (described in the issue #569) has its header whose size is 1f0, but the current computation result is 220, and `Binary::sizeof_headers` returns 400 by `LIEF::align`. So `pe_builder` program produces an invalid PE file from the file.

By removing "+1", `Binary::sizeof_headers` returns 200 properly, and `pe_builder` program produces the same PE file (whose sha256 hash value is 7152c6554215cabb53c75d1cd7d988c10351958fb7d190625e6875e3ce57756e) from the file.

### DIFF
--- a/src/PE/Binary.cpp
+++ b/src/PE/Binary.cpp
@@ -549,8 +549,8 @@ uint32_t Binary::sizeof_headers(void) const {
     size += sizeof(pe64_optional_header);
   }
 
-  size += sizeof(pe_data_directory) * (this->data_directories_.size() + 1);
-  size += sizeof(pe_section) * (this->sections_.size() + 1);
+  size += sizeof(pe_data_directory) * (this->data_directories_.size());
+  size += sizeof(pe_section) * (this->sections_.size());
   size = static_cast<uint32_t>(LIEF::align(size, this->optional_header().file_alignment()));
   return size;
 


romainthomas:
Thanks!

Issue #442, opened 2020-07-20
Lazy loading to improve LIEF parsing performance


Wenzel:
**Describe the bug**
LIEF parsing tend to be very slow on huge binaries.
For example printing whether pandoc binary (118 MB) has NX takes **48 seconds** ! :fearful: 

**To Reproduce**
- `apt-get install pandoc`
- `time python -c "import lief; binary = lief.parse('/usr/bin/pandoc'); print(binary.has_nx)"`

![Capture d’écran de 2020-07-20 22-38-30](https://user-images.githubusercontent.com/964610/87984122-c761a280-cad9-11ea-847b-3bbb6aafdbc6.png)

**Expected behavior**
This operation should take only a few hundred milliseconds to complete.

**Environment (please complete the following information):**
 - System and Version : Ubuntu 20.04
 - Target format ELF
 - LIEF commit version: `0.10.1-bfe5414`

**Additional context**
I'm writing a Python implementation of [checksec.sh](https://github.com/slimm609/checksec.sh) based on LIEF,
and so far the performance is my main issue.

thanks !

jvolkman:
Not sure what @Wenzel's system specs were at the time of testing 2.5 years ago, but I just tried this on Ubuntu 22.04 on an AMD 3970x and it took ~1 second.

```
time python -c "import lief; binary = lief.parse('/usr/bin/pandoc'); print(binary.has_nx)"
True
python -c   1.06s user 0.31s system 99% cpu 1.376 total
```

Wenzel:
@jvolkman thanks for testing again this issue !

Regardless of my machine specs, it seems that there was significant improvment in lief parsing between `0.10.1` and `0.12.3`:
![image](https://user-images.githubusercontent.com/964610/211458414-3825e679-0556-4b27-829b-c9c42d7b2d40.png)

I'll mark this issue as closed for now.

Thank you again !

PR #579, opened 2021-05-18
MachO: Memoize Symbols by address and name during parsing

xhochy:
I was having really long runtimes like also reported in https://github.com/lief-project/LIEF/issues/442 . Actually, my issue is exactly what is covered in https://github.com/conda/conda-build/issues/3684 . Profiling lead to most of the runtime being spent in `Binary::get_symbol` and `BinaryParser::do_bind` (these were actually the only functions visible by eye in the profiling plot). For me the runtime of `lief.parse(…)` goes down as follows for the example of the haskell compiler:

| file | size | time in s (before) | time in s (this PR) |
| --- | --- | -- | -- |
| `libHSbase-4.14.1.0-ghc8.10.4.dylib` | 12M | 47 | 0.2 |
| `libHSghc-8.10.4-ghc8.10.4.dylib` | 97M | 3161 | 1.4 |

### DIFF
--- a/include/LIEF/MachO/BinaryParser.hpp
+++ b/include/LIEF/MachO/BinaryParser.hpp
@@ -20,6 +20,7 @@
 #include <vector>
 #include <limits>
 #include <set>
+#include <map>
 
 #include "LIEF/types.hpp"
 #include "LIEF/visibility.h"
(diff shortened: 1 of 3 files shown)


romainthomas:
Thank you @xhochy for this contribution!

PR #584, opened 2021-06-02
Add str escape

kohnakagawa:
Hi, @romainthomas 

This pull-request fixes minor bugs while calling `to_json` and `ResourcesManager::get_string_file_info` for some PE files.

When non-ascii strings are included in the `lief.to_json` output, the call of `lief.to_json` will crash.
For example, since [this sample](https://www.virustotal.com/gui/file/6a09aef6b70071154e2eba268f5aa3ab342d9e9ecdc6d7a6a5d3c7499d98c5bd/details) contains the non-ascii string in the name filed of symbol, we cannot get the json output of `lief.to_json`. To void this, I added `escape_nonascii` function call to escape non-ascii strings.

Additionally, the current implementation of `ResourcesManager::get_string_file_info` does not catch the exceptions of the `stoul` when passing non-integer string. I added an try-except block to catch these exceptions.

### DIFF
--- a/src/PE/ResourcesManager.cpp
+++ b/src/PE/ResourcesManager.cpp
@@ -608,10 +608,15 @@ ResourceStringFileInfo ResourcesManager::get_string_file_info(const VectorStream
     if (key.length() != 8) {
       LIEF_ERR("Corrupted key ({} {})", u16tou8(key), key_str);
     } else {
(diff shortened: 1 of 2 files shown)


romainthomas:
Thanks @kohnakagawa :+1: 
I merged with https://github.com/lief-project/LIEF/commit/996301623bae3f52ca588e9a4d68e32dec7200ee as I slightly changed  your modification on ``get_string_file_info`` to avoid exceptions.

PR #598, opened 2021-07-28
MachO bug fix: lief.MachO.BindingInfo.library analyze wrong

Kiprey:
There is no need to add `LC_ID_DYLIB` command into `binary_->libraries_`, which can cause the analysis result of `lief.MachO.BindingInfo.library` wrong.

Referring to [llvm /lib/Object/MachOObjectFile.cpp src](https://github.com/llvm-mirror/llvm/blob/2c4ca6832fa6b306ee6a7010bfb80a3f2596f824/lib/Object/MachOObjectFile.cpp#L1401),  it is obvious that LLVM filters the library of LC_ID_DYLIB command.

### DIFF
--- a/src/MachO/BinaryParser.tcc
+++ b/src/MachO/BinaryParser.tcc
@@ -199,7 +199,6 @@ void BinaryParser::parse_load_commands(void) {
       // DyLib Command
       // =============
       case LOAD_COMMAND_TYPES::LC_LOAD_WEAK_DYLIB:
-      case LOAD_COMMAND_TYPES::LC_ID_DYLIB:
       case LOAD_COMMAND_TYPES::LC_LOAD_DYLIB:
       case LOAD_COMMAND_TYPES::LC_REEXPORT_DYLIB:
       case LOAD_COMMAND_TYPES::LC_LOAD_UPWARD_DYLIB:
--- a/src/MachO/Binary.cpp
+++ b/src/MachO/Binary.cpp
@@ -772,8 +772,9 @@ LoadCommand& Binary::add(const LoadCommand& command) {
 
   this->commands_.push_back(copy);
 
-  // Update cache
-  if (typeid(*copy) == typeid(DylibCommand)) {
+  // Update cache  
+  if (typeid(*copy) == typeid(DylibCommand) 
+      && dynamic_cast<DylibCommand*>(copy)->command() != LOAD_COMMAND_TYPES::LC_ID_DYLIB) {
     this->libraries_.push_back(reinterpret_cast<DylibCommand*>(copy));
   }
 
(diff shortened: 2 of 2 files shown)


src/MachO/BinaryParser.tcc
```diff
@@ -199,7 +199,6 @@ void BinaryParser::parse_load_commands(void) {
       // DyLib Command
       // =============
       case LOAD_COMMAND_TYPES::LC_LOAD_WEAK_DYLIB:
-      case LOAD_COMMAND_TYPES::LC_ID_DYLIB:
```
romainthomas, line 202: Why removing this entry from the parser ?

Kiprey: - Similar to the above, there is no need to add ID_DYLIB command to binary_->libraries_. Just similar as the LLVM src, there is also no more processing for LC_ID_DYLIB.
- However, a new container may still be needed to store all LoadCommands for print purpose, but preferably not binary_->libraries_.

romainthomas: I agree with both points but here you break the support of ``LC_ID_DYLIB``

Kiprey: You are right. It is really not good to simply delete a case.

By comparing Binary.cpp and BinaryParser.hpp, I found that Binary.cpp will enable `binary_->commands_` to store all load commands (including LC_ID_DYLIB), but BinaryParser.hpp does not enable the container. 

Maybe it would be better to enable `commands_` to store(support) LC_ID_DYLIB in BinaryParser.hpp? 

I'm not sure.

romainthomas:
Could you attach the binary (or a way to build it) so that we can reproduce the issue?

Kiprey:
> Could you attach the binary (or a way to build it) so that we can reproduce the issue?

Binary link: 
[libsystem_malloc.dylib.zip](https://github.com/lief-project/LIEF/files/6942889/libsystem_malloc.dylib.zip)

It is very easy to reproduce the issue.

just run the script below:

```python
import lief

binary = lief.parse("libsystem_malloc.dylib")

for symbol in binary.imported_symbols:
    if symbol.name == "_printf":
        assert symbol.has_binding_info
        binding_info = symbol.binding_info
        print(binding_info)
```

and the output is 

```text
Class:       LAZY
Type:        POINTER
Address: 0x  1d198
Symbol:      _printf
Segment:     __DATA
Library:     /usr/lib/system/libsystem_platform.dylib
```

However, the real library is `libsystem_c.dylib`. Just as the following output.

```bash
$ nm -u -m libsystem_malloc.dylib | grep "_printf" 
                 (undefined) external _printf (from libsystem_c)
```

PR #608, opened 2021-09-22
Fix DEX prototype parameters type parsing

onlyzdd:
This pull request fixes DEX prototype parameters type parsing bug. Each `param_type` takes 16 bits, not 32 bits. Without this fix, `prototype->params_.size()` is always 0 or 1, which is usually wrong and different from `nb_params`.
See: Dex::TypeList in [kaitai formats: DEX](http://formats.kaitai.io/dex/dex.svg), TYPE LIST in [corkami pics: DEX](https://raw.githubusercontent.com/corkami/pics/master/binary/DEX.png)


romainthomas:
Nice catch! Thanks.

PR #547, opened 2021-02-14
Parse fields in DEX

DanielFi:
Noticed that this is missing while I was using this awesome library, and decided to do it myself!  
Hopefully there's no problem merging this :crossed_fingers:

Complete with:
* python API
* tests
* updated python examples
* json dumping

PR #620, opened 2021-10-17
Fix memory leak in Android DEX parsing

onlyzdd:
This pull request aims to fix the memory leak issue in Android DEX parsing. Memory analysis has been performed with valgrind memcheck tool. 
PS: I can upload the valgrind memcheck log if needed.

### DIFF
--- a/src/DEX/Class.cpp
+++ b/src/DEX/Class.cpp
@@ -243,6 +243,7 @@ std::ostream& operator<<(std::ostream& os, const Class& cls) {
     os << " - " << cls.source_filename();
   }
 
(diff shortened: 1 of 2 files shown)


romainthomas:
Hi @onlyzdd 

My bad, I missed this leak from the PR #547. According to the code, it looks like the ownership of the field is the ``DEX::File`` object so I think it would be better to ``delete`` in this object (or to use a ``std::unique_ptr``)

onlyzdd:
Perfect. Thank you, and I'm going to close this PR.

PR #633, opened 2021-12-08
Walk the max bucket in GNU_HASH for counting dynamic symbols

### LINKED ISSUE #632: Dynamic symbol over-counting using GNU_HASH sections
**Describe the bug**

When using the `LIEF::ELF::DYNSYM_COUNT_METHODS::COUNT_HASH` (or `COUNT_AUTO`, which is default) methods for counting dynamic symbols of an ELF with a `GNU_HASH` section, it may over-count the symbols. This means when using the dynamic symbol iterator, symbols can be read from beyond the end of the `.dynsym` section.

The bug results from not handling bucket elements with index 0; the blog below notes that 0 is a reserved value:

> A bucket element will contain the index 0 if there is no symbol in the hash table for the given value of N. As index 0 of the dynsym is a reserved value, this index cannot occur for a valid symbol, and is therefore non-ambiguous.

https://blogs.oracle.com/solaris/post/gnu-hash-elf-sections

Additionally, looking at the history of this project, the `COUNT_HASH` and `COUNT_RELOCATIONS` options were added to support binaries with no section information. This bug occurs even if section information is present. In my opinion, the `.dynsym` section size should be used to calculate the definitive dynamic symbol count in `AUTO` mode if section information is present, since it is less complex and should be faster and less error-prone.

**To Reproduce**

```cpp
#include <LIEF/LIEF.hpp>

int main(void)
{
    std::string Path = "./sample.elf";

    std::unique_ptr<LIEF::ELF::Binary> ElfSection{LIEF::ELF::Parser::parse(Path, LIEF::ELF::DYNSYM_COUNT_METHODS::COUNT_SECTION)};
    LIEF::ELF::it_symbols SectionSymbolIt = ElfSection->dynamic_symbols();
    std::cout << "Section Count: " << std::distance(std::begin(SectionSymbolIt), std::end(SectionSymbolIt)) << std::endl;

    std::unique_ptr<LIEF::ELF::Binary> ElfHash{LIEF::ELF::Parser::parse(Path, LIEF::ELF::DYNSYM_COUNT_METHODS::COUNT_HASH)};
    LIEF::ELF::it_symbols HashSymbolIt = ElfHash->dynamic_symbols();
    std::cout << "Hash Count: " << std::distance(std::begin(HashSymbolIt), std::end(HashSymbolIt)) << std::endl;
}
```

This requires having a binary with some empty buckets in `GNU_HASH`, and with values such that the loop that checks for the end of hash chains walks beyond the end of the section. I have one that I can share if needed, but it's > 100 MB so too large to attach here.

**Expected behavior**
`COUNT_HASH` and `COUNT_SECTION` should return the same number of symbols; `COUNT_HASH` should not return more symbols than could possibly fit in the `.dynsym` section.

 - System and Version : Ubuntu 20.04
 - Target format: ELF
 - LIEF version: 0.11.5

**Additional context**
I will create a PR with fixes shortly.

adamjseitz:
Resolves #632

### DIFF
--- a/src/ELF/Parser.tcc
+++ b/src/ELF/Parser.tcc
@@ -747,16 +747,15 @@ uint32_t Parser::nb_dynsym_gnu_hash(void) const {
   using Elf_Off  = typename ELF_T::Elf_Off;
 
   const DynamicEntry& dyn_hash = this->binary_->get(DYNAMIC_TAGS::DT_GNU_HASH);
-  const Elf_Off sysv_hash_offset = this->binary_->virtual_address_to_offset(dyn_hash.value());
+  const Elf_Off gnu_hash_offset = this->binary_->virtual_address_to_offset(dyn_hash.value());
 
 
-  this->stream_->setpos(sysv_hash_offset);
+  this->stream_->setpos(gnu_hash_offset);
   if (not this->stream_->can_read<uint32_t>()) {
     return 0;
   }
 
-  const uint32_t nbuckets  = std::min(this->stream_->read_conv<uint32_t>(), Parser::NB_MAX_BUCKETS);
-
+  const uint32_t nbuckets  = this->stream_->read_conv<uint32_t>();
   if (not this->stream_->can_read<uint32_t>()) {
     return 0;
   }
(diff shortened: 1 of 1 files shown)


src/ELF/Parser.tcc
```diff
@@ -522,16 +522,12 @@ uint32_t Parser::get_numberof_dynamic_symbols(DYNSYM_COUNT_METHODS mtd) const {
...
-        nb_dynsym = this->get_numberof_dynamic_symbols<ELF_T>(DYNSYM_COUNT_METHODS::COUNT_RELOCATIONS);
-
-        nb_dynsym_tmp = this->get_numberof_dynamic_symbols<ELF_T>(DYNSYM_COUNT_METHODS::COUNT_SECTION);
-
-        if (nb_dynsym_tmp < Parser::NB_MAX_SYMBOLS and
-            nb_dynsym_tmp > nb_dynsym and
-            (nb_dynsym_tmp - nb_dynsym) < Parser::DELTA_NB_SYMBOLS) {
-          nb_dynsym = nb_dynsym_tmp;
+        nb_dynsym = this->get_numberof_dynamic_symbols<ELF_T>(DYNSYM_COUNT_METHODS::COUNT_SECTION);
+        if (nb_dynsym != 0) {
```
romainthomas, line 526: I disagree with this early return since you can fake the ``.dynsym`` section to falsely reduce the number of symbols (IDA has this issue)

adamjseitz: That makes sense. I did not consider the case where the `.dynsym` section's size would be intentionally shortened to hide symbols. 

I removed this change from the PR, which was independent of the change in `nb_dynsym_gnu_hash` anyway.

PR #645, opened 2021-12-26
Add support for the reserved PE data directory

hypervis0r:
I was doing a proof of concept for an upcoming blog post and noticed that I was unable to access the reserved PE data directory, so I quickly went ahead and added the functionality.


romainthomas:
Could you detail a bit more what is the purpose of this feature. I mean it's ok to have a PoC but why we should integrate your modifications in LIEF?

hypervis0r:
> Could you detail a bit more what is the purpose of this feature. I mean it's ok to have a PoC but why we should integrate your modifications in LIEF?

In my case, I needed a spare data directory to improve lookup times for an added section. I'm sure I'm not alone in wanting to take advantage of this empty data directory for whatever purpose, and I feel like having the ability to read and modify this directory would be a great addition.

Another case is if a malware strain/packer ends up utilizing this reserved data directory, having the ability to access this directory in LIEF would be extremely useful.

include/LIEF/PE/enums.inc
```diff
@@ -225,7 +225,8 @@ enum _LIEF_EN(DATA_DIRECTORY) {
   _LIEF_EI(IAT)                     = 12,
   _LIEF_EI(DELAY_IMPORT_DESCRIPTOR) = 13,
   _LIEF_EI(CLR_RUNTIME_HEADER)      = 14,
-
+  _LIEF_EI(RESERVED)                = 15,
+  
   _LIEF_EI(NUM_DATA_DIRECTORIES)    = 15
```
limitedeternity, line 230: `NUM_DATA_DIRECTORIES` is 16 now, isn't it?

limitedeternity (changes requested):

include/LIEF/PE/enums.inc
hypervis0r: the enum is zero-indexed, as there are 16 total data directories. 0-15 == 16 total

romainthomas:
If I'm not wrong, ``RESERVED`` is basically used to name the a slice of the padding area between the ``DataDirectory`` table and the section table right?

hypervis0r:
> If I'm not wrong, `RESERVED` is basically used to name the a slice of the padding area between the `DataDirectory` table and the section table right?

Yes, as seen [here in the MSDN PE format page](https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#optional-header-data-directories-image-only)

romainthomas:
Ok so I think you PR is not the best approach for the following reasons:

1. What if a PE does not have this padding area?
2. If we follow your logic, we should have ``RESERVED1``, ``RESERVED2``, ...
3. ``RESERVED`` does not really exist in the PE format and it could be confusing.

I suggest you to make an API over: https://github.com/lief-project/LIEF/blob/0459d3bf777ae9f99ad08cfd7f341992b0debdb4/include/LIEF/PE/Binary.hpp#L482

hypervis0r:
I disagree with the fact that this `RESERVED` data directory does not exist in the PE format. As shown in the MSDN page for the PE format, there is a clear reserved directory listed in the list of data directories.

image

In addition to the MSDN page regarding the PE format, the official Microsoft headers regarding PE files even shows the data directory array to be a static size of 16. If this reserved data directory was "optional" in any way, it would break the optional header of the NT header.

```cpp
// located in winnt.h
#define IMAGE_NUMBEROF_DIRECTORY_ENTRIES    16

typedef struct _IMAGE_OPTIONAL_HEADER64
{
    // [...snip...]
    IMAGE_DATA_DIRECTORY DataDirectory[IMAGE_NUMBEROF_DIRECTORY_ENTRIES];
} IMAGE_OPTIONAL_HEADER64, *PIMAGE_OPTIONAL_HEADER64;
```

romainthomas:
Ok my bad.

limitedeternity:
@hypervis0r, I think it'd be great if you could provide a test-case.

romainthomas:
Merged with https://github.com/lief-project/LIEF/commit/e12dafbfc0ac33231f35326707a23e68ca8dc89e

PR #667, opened 2022-02-15
Fix external span usage

ekilmer:
* Typo in the definition check

* Search for <tcb/span.hpp> instead of just <span.hpp>


romainthomas:
Thanks!

PR #669, opened 2022-02-21
Use CMake BUILD_INTERFACE generator expression for include directories

ekilmer:
This fixes the paths that are generated for the exported targets so that
they do not have absolute paths that were found/used at build time.

Note: It doesn't look like the generated `LIEFExport.cmake` file is
actually used anywhere, so these changes don't really have any effect
yet, but I am hoping to fix that in an upcoming PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -336,7 +336,7 @@ configure_file(${CMAKE_CURRENT_SOURCE_DIR}/LIEF.pc.in ${CMAKE_CURRENT_BINARY_DIR
                @ONLY)
 
 target_include_directories(LIB_LIEF
-  PUBLIC  "${LIEF_PUBLIC_INCLUDE_DIR}"
+  PUBLIC  "$<BUILD_INTERFACE:${LIEF_PUBLIC_INCLUDE_DIR}>"
   PRIVATE "${LIEF_PRIVATE_INCLUDE_DIR}")
 
 if(LIEF_ENABLE_JSON)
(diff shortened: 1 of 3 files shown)


romainthomas:
Hi @ekilmer 
I use sometimes ``LIEFExport.cmake`` to quickly link against a built version of LIEF in other projects so it would be nice to keep it working :)

ekilmer:
> Hi @ekilmer I use sometimes `LIEFExport.cmake` to quickly link against a built version of LIEF in other projects so it would be nice to keep it working :)

Understood. This PR shouldn't affect the currently generated `LIEFExport.cmake` because it is generated through CMake's [`export` command](https://cmake.org/cmake/help/v3.5/command/export.html), which is for targets in the build directory and thus the `BUILD_INTERFACE` generator expression will be turned on, and these include paths will still appear correctly.

That is opposed to CMake's [`install(EXPORT` command](https://cmake.org/cmake/help/v3.5/command/install.html#installing-exports), which is not currently included in this project (yet), but can be used for linking against targets that are installed, and the `BUILD_INTERFACE` will be turned _off_ because build-time directories don't make sense when something is installed.

Give me a day or two to test the intended fully working solution that supports both using the build directory targets and installed targets while letting CMake do a lot of the work for handling the different target properties like library locations, include directories, and dependency linking. The goal is to replace the manual creation of the `LIEF::LIEF` target in the `FindLIEF.cmake` file with just an `include(LIEFExport.cmake)` https://github.com/lief-project/LIEF/blob/566d4ac8a86e458a40de1e2816115f8db48a5501/scripts/FindLIEF.cmake#L154-L170

romainthomas:
Thank you @ekilmer for the description, I'll take a closer look at the PR this week-end :)

ekilmer:
@romainthomas Took a little longer than I thought after realizing you wanted to support an installation that has both static and shared libraries, but I believe the changes in #674 accomplish that. I'm very happy to discuss the changes and would really work hard to get it merged. Thank you for the wonderful project!

PR #670, opened 2022-02-24
support for adding and assigning new auxiliary symbol versions

ahaensler:
This PR has helped me patch an old library. It was built against an old glibc version, and ``pthread_yield`` didn't have an auxiliary symbol version entry. I made it work by adding "GLIBC_2.2".

The use case in Python

```
import lief

filename = "library.so"
lib = lief.parse(filename)

for svr in lib.symbols_version_requirement:
    if svr.name == "libc.so.6":
        ar = lief.ELF.SymbolVersionAuxRequirement()
        ar.other = 12
        ar.name = "GLIBC_2.2"
        sauxr = svr.add_auxiliary_requirement(ar)
        break

for x in lib.dynamic_symbols:
    if x.name == "pthread_yield":
        x.symbol_version.symbol_version_auxiliary = sauxr

lib.write(filename)  
```

### DIFF
--- a/api/python/ELF/objects/pySymbolVersion.cpp
+++ b/api/python/ELF/objects/pySymbolVersion.cpp
@@ -18,6 +18,7 @@
 #include "LIEF/ELF/SymbolVersion.hpp"
 #include "LIEF/ELF/hash.hpp"
 #include "LIEF/ELF/SymbolVersionAux.hpp"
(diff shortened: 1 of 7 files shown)


romainthomas:
Thank you @ahaensler!

I merged your commit with a few modifications (documentation, ``const-ness``) here: https://github.com/lief-project/LIEF/commit/9331982bfe963ee6bee91890cfa81c5e35b94bc9

PR #677, opened 2022-03-19
Reformat project using clang-format.

ahawad:
Hello.

This commit adds a clang-format configuration file and reformats the repository with it.

The command I ran to generate the diff is: 

`clang-format **/*.{tcc,.hpp,.h,.cpp,.c}`

I understand that the diff is large because it touches every C++ & C file in the repo, if you'd like to verify that the diff does nothing beyond reformat the files, please do so.

The value from this is that future contributions can be formatted according to a well-defined specification. I am happy to adjust the file if you'd prefer different settings.


ahawad:
I am also working on removing exceptions from the library on top of this change. That commit is incomplete, but I should be done with it over the next couple of days.

romainthomas:
Hi @ahawad 
Thank for your contribution. I would suggest you to read this https://lief-project.github.io/blog/2022-02-13-lief-rtti-exceptions/ :)

ahawad:
Hi, @romainthomas.

Thank you for the prompt reply. I have indeed read that blogpost. I am using the result<T> abstraction to replace exceptions. You can see the work in progress in the commit. 

I have tried to also abstract over the implementation of result<T> a bit more, by treating as an opaque object in my changes. This has the advantage of making the implementation easy to replace, but it does make the code a bit more verbose. 

I do have a few concerns, mostly surrounding when to fail parsing. The current parser basically never fails. It will keep trudging along no matter the error it encounters. I am preserving that logic by emitting errors when the parser encounters them, but it seems somewhat counter intuitive to have a function able to return errors, but the implementation does not actually propagate the errors up.

The latest iteration of the commit has most of the work done for ELF. There is one exception left for me to remove in the DynamicEntryArray.cpp file. This leaves the ones in the other formats, but the other formats don't use exceptions as heavily as ELF does. I believe I should be done with this today/tomorrow. 

Do you have any concerns other than ones raised in the blogpost?

romainthomas:
Actually I welcome contributions but for this internal parts, I prefer to handle these changes by myself. 
For the clang-format, I completely agree on its purpose but I would welcome a configuration that matches as close as possible the current formatting.

ahawad:
No worries. I will close the PR, but I will keep the fork. For the formatting, I am not really sure what the current formatting is. I am happy to change the configuration to something different, but I am not sure what configuration would get us there.

FWIW, most of the exception-related changes are fairly mechanical, because the idiom change was largely adopted within the library. I would suspect that you'll find a lot of the work in the commit is not too different from you might end up doing for v0.13.0.

PR #692, opened 2022-04-01
Fix ELF section inclusion calculations

kjackiewicz:
ELF section's virtual address was compared with segment physical/file span. It works if section's virtual address is equal to its file offset which is most often true. If it's not, the section may get additional invalid entries on associated segments list. Fixed.

Also, the condition comparing physical/file span ends seems to be missing an '='. Added.


romainthomas:
Thank you for spotting this error!

PR #695, opened 2022-04-08
Fix project using deprecated iterator base class

### LINKED ISSUE #694: Project use the deprecated iterator base class
Current [iterators.hpp](https://github.com/lief-project/LIEF/blob/master/include/LIEF/iterators.hpp) uses the [std::iterator](https://en.cppreference.com/w/cpp/iterator/iterator) base class that was deprecated in C++17 and is generating deprecated warnings in the newer compilers.

I confirmed the issue with the Apple clang version 13.1.6.

https://github.com/lief-project/LIEF/blob/57647bb05b7f1c0054e959a10faa9c3687cdda05/include/LIEF/iterators.hpp#L46-L54

https://github.com/lief-project/LIEF/blob/57647bb05b7f1c0054e959a10faa9c3687cdda05/include/LIEF/iterators.hpp#L263-L272

A safe way of refactoring this header is by defining the expected types as described in the original language issue entry that led to the deprecation: https://timsong-cpp.github.io/lwg-issues/2438

My proposed fix: #695

abraaocsantana:
Fixes #694

### DIFF
--- a/include/LIEF/iterators.hpp
+++ b/include/LIEF/iterators.hpp
@@ -46,13 +46,14 @@ using add_lvalue_reference_t = typename std::add_lvalue_reference<T>::type;
 //! Iterator which returns reference on container's values
 template<class T, typename U = typename decay_t<T>::value_type,
          class ITERATOR_T = typename decay_t<T>::iterator>
-class ref_iterator : public std::iterator<
-                     std::bidirectional_iterator_tag,
-                     decay_t<U>,
-                     ptrdiff_t,
-                     typename std::remove_pointer<U>::type*,
-                     typename std::remove_pointer<U>::type&> {
+class ref_iterator {
   public:
+  using iterator_category = std::bidirectional_iterator_tag;
+  using value_type = decay_t<U>;
+  using difference_type = ptrdiff_t;
+  using pointer = typename std::remove_pointer<U>::type*;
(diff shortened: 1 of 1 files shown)


abraaocsantana:
Copy of the original text of the Issue, for a more straightforward evaluation:

> Current [iterators.hpp](https://github.com/lief-project/LIEF/blob/master/include/LIEF/iterators.hpp) uses the [std::iterator](https://en.cppreference.com/w/cpp/iterator/iterator) base class that was deprecated in C++17 and is generating deprecated warnings in the newer compilers.
> 
> I confirmed the issue with the Apple clang version 13.1.6.
> 
> https://github.com/lief-project/LIEF/blob/57647bb05b7f1c0054e959a10faa9c3687cdda05/include/LIEF/iterators.hpp#L46-L54
> 
> https://github.com/lief-project/LIEF/blob/57647bb05b7f1c0054e959a10faa9c3687cdda05/include/LIEF/iterators.hpp#L263-L272
> 
> A safe way of refactoring this header is by defining the expected types as described in the original language issue entry that led to the deprecation: https://timsong-cpp.github.io/lwg-issues/2438
> 
> My proposed fix: #695

romainthomas:
Thank you for the fix. It's good with me.

PR #698, opened 2022-04-20
using CMAKE_CURRENT_SOURCE_DIR in place of CMAKE_SOURCE_DIR for Fetch…

pive:
When using FetchContent to use LIEF from within a repo, cmake fails in `LIEFGit.cmake` because git commands are made from within the "host" repo (`CMAKE_SOURCE_DIR`) instead of the LIEF repo. This issue doesn't occur in `examples/cmake/add_subdirectory` because the main `CMakeLists.txt` is within LIEF repo. This is solved by using `CMAKE_CURRENT_SOURCE_DIR` instead of `CMAKE_SOURCE_DIR` in `LIEFGit.cmake` (https://cmake.org/cmake/help/latest/variable/CMAKE_CURRENT_SOURCE_DIR.html) as working directory for `git` commands.


romainthomas:
Thank you!

Issue #705, opened 2022-05-16
LIEF is aborting its client application when dealing with a specific malware binary


rafael-santiago:
**Describe the bug**
When ``LIEF::PE::Binary::get_content_from_virtual_address()`` is called it is aborting the LIEF's client application. However, it only occurs when trying to parse a specific malware ([see](https://www.virustotal.com/gui/file/6df08107847e05fb150b0aa3cd009daaf2e352302245f76e54225587654adf92) virustotal entry).

According to MSVC's error report, the bug seems to be related to some invalid memory access, maybe an iterator exceding its limits etc.

**To Reproduce**
Build and execute the PoC **into a sandbox**.

**Expected behavior**
Even not being able to parse the malicious PE it should throw some exception but it should not abort its client application.

**Environment (please complete the following information):**
 - System and Version : ``Windows 10`` (``LIEF`` was built by using ``MSVC 2019``)
 - Target format : ``PE``
 - LIEF commit version: There is no version information inside my ``LIEF/config.h``  but I believe that it occurs with the ``upstream``, too. I have been testing it with pretty old version (when the method ``has_signature()`` was still available) and also with a version built from ``upstream`` last month.

**Additional context**
None.

**Detailed issue related by MSVC instrumented code**
![LIEFBug](https://user-images.githubusercontent.com/7308777/168691535-b9c214c9-7e10-4ed5-9e69-136f1221c919.png)

**Poc source code**
[LIEFBug.cc.txt](https://github.com/lief-project/LIEF/files/8703768/LIEFBug.cc.txt)

PR #707, opened 2022-05-19
Detect data section end's before going ahead on accessing it

rafael-santiago:
Hi there!

This commit is related to this specific [issue](https://github.com/lief-project/LIEF/issues/705). I have obeserved that the adjustment done under the changed if-block was overflowing the unsigned variable ``checked_size`` (it was loaded with a value less than the delta subtracted from it). Being an overflowed value it was making LIEF read more data than it supposed must do. Since I am not totally engaged with the code I decided to avoid the problem by inserting a defensive escape. Anyway, the following problem seems to be occurring with small binaries. Since LIEF is used by security applications it would be good avoid this kind of problem, since it can take the client application down. Checking on the sanity of pointers addresses before accessing them also would be a good defensive measure for the next releases.

I was not able to reproduce the [problem](https://github.com/lief-project/LIEF/issues/705) with the LIEF generated with this proposed patch. You can use the PoC attached to the issue in order to atest it. 

I hope that helps!

### DIFF
--- a/src/PE/Binary.cpp
+++ b/src/PE/Binary.cpp
@@ -1449,7 +1449,12 @@ std::vector<uint8_t> Binary::get_content_from_virtual_address(uint64_t virtual_a
   const uint64_t offset = rva - section->virtual_address();
   uint64_t checked_size = size;
   if ((offset + checked_size) > content.size()) {
-    checked_size = checked_size - (offset + checked_size - content.size());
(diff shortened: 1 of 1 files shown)


romainthomas:
Sorry for the delay. Thank you for the fix!

rafael-santiago:
You are welcome!

PR #716, opened 2022-06-11
Add API to remove a section from a segment & section names

dsanders11:
Since two Mach-O sections could have the same name, an API to remove a section in the specified segment name is useful.


romainthomas:
Perfect PR. Nothing to say :) Thank you!

PR #718, opened 2022-06-13
Correct default for Binary::has_nx on ppc64

### LINKED ISSUE #717: ELF has_nx is not correct on ppc64
On 64-bit PowerPC, the default if there is no `GNU_STACK` program header is-no-exec.

```
/*
 * This is the default if a program doesn't have a PT_GNU_STACK
 * program header entry. The PPC64 ELF ABI has a non executable stack
 * stack by default, so in the absence of a PT_GNU_STACK program header
 * we turn execute permission off.
 */
```
— https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/powerpc/include/asm/page_64.h#n91

However,  LIEF hardcodes a default of false for `has_nx` here: https://github.com/lief-project/LIEF/blob/master/src/ELF/Binary.cpp#L1100 . This causes it to misreport the stack as executable in the default case.

Edit: I don't know if this specific to Linux ELF or not. I guess so. Then again, so is `PT_GNU_STACK` I suppose.

laanwj:
From the Linux kernel source:

     * This is the default if a program doesn't have a PT_GNU_STACK
     * program header entry. The PPC64 ELF ABI has a non executable stack
     * stack by default, so in the absence of a PT_GNU_STACK program header
     * we turn execute permission off.

Resolve #717

### DIFF
--- a/src/ELF/Binary.cpp
+++ b/src/ELF/Binary.cpp
@@ -1097,7 +1097,12 @@ bool Binary::has_nx() const {
                                        return segment->type() == SEGMENT_TYPES::PT_GNU_STACK;
                                      });
   if (it_stack == std::end(segments_)) {
-    return false;
+    if (header().machine_type() == ARCH::EM_PPC64) {
+      // The PPC64 ELF ABI has a non-executable stack by default.
+      return true;
+    } else {
+      return false;
+    }
   }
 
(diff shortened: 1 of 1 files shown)


romainthomas:
Great thank you!

PR #723, opened 2022-06-15
Fix ordering of resource node childs in PE

dsanders11:
This fixes the ordering of resource node childs, drops `lief.PE.ResourceNode.sort_by_id` (not needed since sorting is done on add), and adds test coverage.

Implemented to this description in [some documentation on Microsoft's site](https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#resource-directory-entries) (emphasis on the relevant portion):

> The directory entries make up the rows of a table. Each resource directory entry has the following format. Whether the entry is a Name or ID entry is indicated by the resource directory table, which indicates how many Name and ID entries follow it **(remember that all the Name entries precede all the ID entries for the table). All entries for the table are sorted in ascending order: the Name entries by case-sensitive string and the ID entries by numeric value.** Offsets are relative to the address in the IMAGE_DIRECTORY_ENTRY_RESOURCE DataDirectory. See [Peering Inside the PE: A Tour of the Win32 Portable Executable File Format](https://docs.microsoft.com/en-us/previous-versions/ms809762(v=msdn.10)#pe-file-resources) for more information.

### DIFF
--- a/api/python/PE/objects/pyResourceNode.cpp
+++ b/api/python/PE/objects/pyResourceNode.cpp
@@ -95,10 +95,6 @@ void create<ResourceNode>(py::module& m) {
         "Delete the " RST_CLASS_REF(lief.PE.ResourceNode) " with the given :attr:`~lief.PE.ResourceNode.id` from childs",
         "id"_a)
 
-    .def("sort_by_id",
-        &ResourceNode::sort_by_id,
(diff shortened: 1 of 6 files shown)


romainthomas:
Great! Thank you.

romainthomas:
Great! Thank you.

PR #736, opened 2022-06-22
Pass jobs value to ninja as well

robertgzr:
Signed-off-by: Robert Günzler <r@gnzler.io>


romainthomas:
I identified a regression with this PR. By default Ninja uses all the threads available for compiling. With this commit, it requires to **manually provide** the number of threads.
Therefore, I updated your commit to match the previous behavior: d1ab0c71f55ca71453b604ded26ae47873fc05ce

Issue #749, opened 2022-07-15
ELF symbol versions: `SymbolVersionDefinition` with multiple entries do not get correctly linked to their `SymbolVersion`


aeflores:
A `SymbolVersionDefinition` can contain multiple `SymbolVersionAux` entries. However, a `SymbolVersion` object can only hold one `SymbolVersionAux`. As a result, the code that assigns the `SymbolVersionAux` from a `SymbolVersionDefinition` to the corresponding `SymbolVersion` repeatedly writes on the same field https://github.com/lief-project/LIEF/blob/master/src/ELF/Parser.tcc#L1530 and only the last entry in the `SymbolVersionDefinition` is associated with the symbol version.

**To Reproduce**

You can reproduce the bug with the following example:

foo.c
```
#include <stdio.h>

__asm__(".symver foo2,foo@LIBFOO_2.0");
void foo2(void)
{
    puts("foo 2.0");
}
```
Compiled with linker script `libfoo.map`
```
LIBFOO_1.0 {
    global:
      foo;
    local:
      *;
};

LIBFOO_2.0 {
    global:
      foo;
    local:
      *;
} LIBFOO_1.0;
```
The compilation command would be:
```
gcc foo.c -o libfoo.so -shared  -fPIC -Wl,--version-script=libfoo.map
```
Then we can load the library with lief (python API)
```
>>> foo_bin= lief.parse("libfoo.so")
>>> foo_sym = [sym for sym in foo_bin.dynamic_symbols if sym.name == "foo"][0]
>>> foo_sym.symbol_version.symbol_version_auxiliary.name
'LIBFOO_1.0'
```

What is happening is that the symbol `foo` is associated with a `SymbolVersionDefinition` that has two entries `LIBFOO_2.0` and `LIBFOO_1.0`. Since the second entry is `LIBFOO_1.0`, that is the one that get associated to the symbol version in the end.

Note that other tools will report the first entry instead of the last one. This is the case of `readelf` e.g.:
```
>> readelf libfoo.so -aW | grep foo
6: 000000000000060a    19 FUNC    GLOBAL DEFAULT   13 foo@LIBFOO_2.0
```

**Expected behavior**

I think the solution is for `SymbolVersion` to be able to point to a `SymbolVersionDefinition` that might have multiple entries or to contain a list of `SymbolVersionAux` directly. I suspect a similar issue can happen with `SymbolVersionRequirement` since those can have multiple entries as well. This of course would be a breaking change since it would change the public API.

[...]

PR #751, opened 2022-07-19
Only assign a SymbolVersionAux to a SymbolVersion if it does not have one

aeflores:
This makes sure that whenever we have a SymbolVersionDefinition with multiple
entries only the first entry gets associated to the SymbolVersion.
Subsequent entries represent version predecessors, not the main version of the symbol
definition. See issue https://github.com/lief-project/LIEF/issues/749


aeflores:
What is the best way to add the test case from https://github.com/lief-project/LIEF/issues/749? Where would it belong?

romainthomas:
Merged with 05e52e11bf0153c3ef2a2a451be5a7782646693c

PR #752, opened 2022-07-21
Add overloads to write output to std::ostream

ZehMatt:
This allows the user to write to anything that inherits from std::ostream. This is also a work-around for Windows as std::string typically holds ascii and not utf8 so using unicode paths is not particular well supported, with this PR the user can now open an fstream with either wstring or u8string (C++20) and simply pass the stream object. This also allows to write to a stream such std::stringstream allowing to work only with memory.


romainthomas:
Perfect! I updated the changelog and I merged your commits through 9d55f538602989c69454639565910884c5c5ac7c

PR #754, opened 2022-07-22
Fix some compiler warnings

ZehMatt:


romainthomas:
Great! Waiting for the CI check and I merge your commits

ZehMatt:
Some tests seem to be failing but I'm not quite sure why.

romainthomas:
I think I know. I'll investigate them tomorrow

ZehMatt:
I suspect the hash function always had a truncated value of 32 bit, now that I fixed that the values no longer match probably, thats just a guess at the moment as I'm not entirely sure if I read the output log correctly.

src/visitors/hash.cpp
```diff
@@ -153,13 +153,13 @@ size_t Hash::hash(const std::vector<uint8_t>& raw) {
 size_t Hash::hash(const void* raw, size_t size) {
   const auto* start = reinterpret_cast<const uint8_t*>(raw);
 
-  std::vector<uint8_t> sha256(32, 0);
+  std::vector<uint8_t> sha256(32, 0u);
   mbedtls_sha256(start, size, sha256.data(), 0);
 
-  return std::accumulate(std::begin(sha256), std::end(sha256), 0,
-     [] (size_t v, uint8_t n) {
+  return std::accumulate(std::begin(sha256), std::end(sha256), 0u,
```
romainthomas, line 159: The purpose of this hashing function is to be on the size of a pointer according to the architecture.
I pushed an update on this branch: [fix/warnings](https://github.com/lief-project/LIEF/tree/fix/warnings).
In particular this commit: https://github.com/lief-project/LIEF/commit/f29aac5ada1e423c46e798f9b98d1b60f69a0db5 Let me know if you agree.

ZehMatt: Makes sense

romainthomas:
Merged :)

PR #758, opened 2022-07-28
Avoid using -1u constant and fix conflict with max() macro from Windows API

rafael-santiago:
This commit introduces one improvement to make easy to build softwares that depends on `LIEF` when treating warnings as error and also a fix to a conflict between `std::numeric_limits::max()` and `max()` function from `Windows API`. If a `LIEF`'s client includes parts of `Windows` headers that will hit on `max` macro and after including some `LIEF` headers, some compilation errors complaining about this conflict will show up.

The improvement related to silent up the `MSVC` "unary minus operator applied to unsigned type, result still unsigned" is about stop using the hardcoded `-1u` and start using the suitable constants defined in `climits` (i.e.: `limits.h`). Being `UINT_MAX` for `uint32_t` and `SIZE_MAX` for `size_t`.

The fix related to the compilation error produced by the conflict with `max()` macro from `Windows API` was merely wrap
`std::numeric_limits::max()` statement with `(...)`.

I hope that helps!

### DIFF
--- a/include/LIEF/DEX/Class.hpp
+++ b/include/LIEF/DEX/Class.hpp
@@ -16,6 +16,8 @@
 #ifndef LIEF_DEX_CLASS_H_
 #define LIEF_DEX_CLASS_H_
 
+#include <climits>
+
 #include "LIEF/visibility.h"
 #include "LIEF/Object.hpp"
 #include "LIEF/iterators.hpp"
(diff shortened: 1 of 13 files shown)


romainthomas:
Yes, it's better. Thank you!

rafael-santiago:
> Yes, it's better. Thank you!

You are welcome!

PR #780, opened 2022-09-10
Fix some conversions of 64-bit to 32-bit types in MachO code

dsanders11:
Discovered these when compiling LIEF with Emscripten, which is a 32-bit environment. There's various places where `uint64_t` values are assigned to a `size_t`, so when `size_t` is 32-bit, this causes a conversion which might truncate values. I specifically saw it affecting the virtual address for sections.

This is by no means exhaustive, I only changed enough to fix the issue I was seeing and a few others I found. I expect there might be follow-up PRs doing more of this kind of fixes. Probably would be a good idea to enable `-Wshorten-64-to-32` to catch these, but at the moment enabling that produces hundreds of warnings, so more cleanup needs to be done first.

### DIFF
--- a/include/LIEF/MachO/Binary.hpp
+++ b/include/LIEF/MachO/Binary.hpp
@@ -667,7 +667,7 @@ class LIEF_API Binary : public LIEF::Binary  {
   //! Default constructor
   Binary();
 
-  void shift_command(size_t width, size_t from_offset);
+  void shift_command(size_t width, uint64_t from_offset);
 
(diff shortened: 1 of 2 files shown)


dsanders11:
Not sure why GitHub thinks I'm a first-time contributor and won't run CI in this case. 🤔 I haven't done a full test suite run with these changes so want to ensure the CI run is green.

romainthomas:
Ok, it should run now

dsanders11:
@romainthomas, looks like tests all pass.

PR #798, opened 2022-09-23
chore: Add gitignore

its0x08:
- Create `.gitignore` file to avoid overflowing VSCode with changed, that belong to the building process, after having built the project.


romainthomas:
Thank you @its0x08 :+1:

PR #807, opened 2022-10-09
Updates the .dynamic section's contents after rebuilding it.

ahawad:
Hello, 

I found what I think is a bug with LIEF's ELF builder. It seems that the builder does not update the dynamic section's contents after rebuilding it. I am not sure if the intention of the code is to update the section indirectly by updating the segment's contents, but as far as I can tell, that does not always work.

### DIFF
--- a/src/ELF/Builder.tcc
+++ b/src/ELF/Builder.tcc
@@ -43,7 +43,6 @@
 #include "LIEF/ELF/SymbolVersionAuxRequirement.hpp"
 #include "LIEF/ELF/Note.hpp"
 
-#include "LIEF/ELF/Builder.hpp"
 #include "LIEF/errors.hpp"
 
 #include "ELF/Structures.hpp"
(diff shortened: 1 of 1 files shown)


romainthomas:
Hello @ahawad 

Actually the section should be updated as it is mirrored with the segment's content. Could you please attach a script that reproduces the fail?

ahawad:
```python
import lief


library = "ELF64_x86-64_library_libtriton.so"

lib = lief.parse(library)
dynamic_entry = lief.ELF.DynamicEntry()
dynamic_entry.tag = lief.ELF.DYNAMIC_TAGS.BIND_NOW
dynamic_entry.value = 1
lib.add(dynamic_entry)
lib.write("temp.out")
lib = lief.parse("temp.out")
dyn_section = lib.get_section(".dynamic")
dyn_segment = lib[lief.ELF.SEGMENT_TYPES.DYNAMIC]

assert dyn_section.size == dyn_segment.virtual_size
```

This assertion fails without the code change in the commit here.

I found the root cause, when one calls `segment.content(new_content)`, there are two code paths. One where the content is part of a `DataHandler` shared with sections and one where the content is in a "cache." If the data is in cache, then behavior assuming the segment and section share an underlying buffer doesn't work. 

However, in the snippet above, both the section and segment share a `DataHandler`, yet it doesn't work. The reason is subtle, but it's because when `segment.content(new_content)` updates a content, it only updates the `Node` with `SEGMENT_TYPE`, but the corresponding section's node doesn't get updated, and consequently, the section's size remains the same as before extending the segment.

ahawad:
I should have been clearer. I think the content will look right, but the section's size will be incorrect.

romainthomas:
Ok got it and **really** thank you for the details about the root cause. I'll check to fix it properly

PR #809, opened 2022-10-12
Bugfix for ImportEntry to accept PE_TYPE

redspot:
In the Python bindings, a new instance of ImportEntry has it's type hard-coded to PE_TYPE::PE32, with no way to change the type once created.

This prevents adding new ordinal imports to PE32+, aka 64-bit, binaries.

The fix is to allow ImportEntry to also accept PE_TYPE. In addition, the unittest also documents how to add an ordinal using LIEF.

* Update ImportEntry to accept PE_TYPE
* Update python bindings for ImportEntry
* Add unittest for ImportEntry and ordinals


romainthomas:
Thank you for this PR, I triggered the CI

PR #815, opened 2022-10-24
Fix name lookup issue on older gcc

toge:
I am trying to update the LIEF conan package.
https://github.com/conan-io/conan-center-index/pull/13660

Older gcc (e.g. gcc 5) seems to have some compile errors due to the name lookup logic.
I have created a PR that adds `this->` as a prefix.
If it could be merged, it would be helpful for future LIEF packaging.


romainthomas:
Done!

PR #819, opened 2022-11-06
Fixes to tutorials

mgunyho:
Hi, I was following along the tutorials and noticed that the tutorial 04 is a bit incomplete. This PR updates the tutorial so that it contains all of the the example Python code [here](https://github.com/lief-project/tutorials/tree/master/04_ELF_hooking) (although note that the hooked function in the tutorial is `exp`, while in the example script it is `cos`). I also updated the address calculation logic in the tutorial to match the example, but I haven't checked if it's correct.

I also fixed a broken link and some small typos.


romainthomas:
Thank you for your review and the associated PR, it has been a while since a didn't review this part so thank you!

PR #823, opened 2022-11-17
Fix ambiguous name error with `using namespace std;`

mrexodia:
When including LIEF after a `using namespace std` you get errors that those helper types are conflicting with the ones in `std`. We moved them inside the `Object` to prevent this error.

Related: https://github.com/lifting-bits/sleigh/pull/143


romainthomas:
Make sense. Thank you.

PR #831, opened 2022-12-02
Avoid memory leak for Dynamic entries in C API

aeflores:
The array and flag fields of the corresponding dynamic entries need to be freed as well.


romainthomas:
Great! Thank you.

Issue #843, opened 2022-12-13
Unmatched size of dynamic table entry tags of C API


liyansong2018:
**Description**
When I tested elf_reader, I found a strange problem: when LEIF analyzes elf dynamic entry, it always tries to read the data at memory address 0x20. I spent some time studying this problem.

**To Reproduce**

construct ELF file with entry tag larger than 32bit (e.g. 0xf0ff0000001a), as shown below.
[poc.zip](https://github.com/lief-project/LIEF/files/10216041/poc.zip)

```shell
┌──(lys㉿kali)-[~/LIEF/build]
└─$ readelf -d ~/id00    
readelf: Warning: Section 11 has an out of range sh_link value of 66
readelf: Warning: Section 27 has an out of range sh_link value of 48
readelf: Error: Reading 12884901905 bytes extends past end of file for string table
readelf: Error: no .dynamic section in the dynamic segment

Dynamic section at offset 0x2df8 contains 26 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x000000000000000c (INIT)               0x1000
 0x000000000000000d (FINI)               0x1158
 0x0000000000000019 (INIT_ARRAY)         0x3de8
 0x000000000000001b (INIT_ARRAYSZ)       8 (bytes)
 0x0000f0ff0000001a (<unknown>: f0ff0000001a)    0x3df0
 0x000000000000001c (FINI_ARRAYSZ)       8 (bytes)
 0x000000006ffffe13 (<unknown>: 6ffffe13) 0x3a0
 0x0000000000000005 (STRTAB)             0x470
 0x0000000000000006 (SYMTAB)             0x3c8
 0x000000000000000a (STRSZ)              141 (bytes)
 0x000000000000000b (SYMENT)             24 (bytes)
 0x0000000000000015 (DEBUG)              0x0
 0x0000000000000003 (PLTGOT)             0x4000
 0x0000000000000002 (PLTRELSZ)           24 (bytes)
 0x0000000000000014 (PLTREL)             RELA
 0x0000000000000017 (JMPREL)             0x600
 0x0000000000000007 (RELA)               0x540
 0x0000000000000008 (RELASZ)             192 (bytes)
 0x0000000000000009 (RELAENT)            24 (bytes)
 0x000000006ffffffb (FLAGS_1)            Flags: PIE
 0x000000006ffffffe (VERNEED)            0x510
[...]

PR #844, opened 2022-12-13
Marks size of dynamic table entry tags in c api

liyansong2018:
Fix https://github.com/lief-project/LIEF/issues/843


romainthomas:
Fixed with c0da465f74630d43e5e6e1095e4f6d2761878141

PR #856, opened 2023-01-17
CMake config: check whether variable is DEFINED

ekilmer:
When running CMake with `-Werror=dev` on an external project that uses/finds LIEF, there is an error if the user did not define the `LIEF_INCLUDE_DIR` variable. This variable is not defined by default.

This commit adds a check for whether the variable is defined.


romainthomas:
Thanks @ekilmer!

PR #876, opened 2023-03-20
Add PE exception info

ZhouZiY:
I add [exception info for PE64](https://learn.microsoft.com/en-us/cpp/build/exception-handling-x64?view=msvc-170#chained-unwind-info-structures). I am not familiar with LIEF and if there are any problems, plz let me know.


romainthomas:
Thank you for proposing this update in LIEF, I'll handle the integration/test suite.

PR #879, opened 2023-03-23
Add LIEF_INSTALL option

nsurbay:
I try to include LIEF with ``FetchContent`` in a project that already depend on ``spdlog``. 
The option ``LIEF_EXTERNAL_SPDLOG`` is set to ON to force lief to use the existing dependencies, however, this generate the followed error :

```
CMake Error: install(EXPORT "LIEFExport" ...) includes target "lief_spdlog" which requires target "spdlog" that is not in any export set.
```

This merge request add an option ``LIEF_INSTALL`` that may be disable to skip the install and package part of LIEF cmake.
This is inspired by the ``SPDLOG_INSTALL`` without the detection if LIEF is the current main project (see [spdlog cmakefile](https://github.com/gabime/spdlog/blob/040874224bff7d61a09cbdb22fa318cbaf05afd7/CMakeLists.txt#L89) )

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -637,18 +637,6 @@ if(LIEF_DOC)
   add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/doc)
 endif()
 
-# Find Package Config
-# ======================
-configure_file(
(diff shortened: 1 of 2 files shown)


romainthomas:
Merci pour le fix Nicolas :)

PR #883, opened 2023-03-29
ELF: fix assertion failure in `Binary::strings()`

nikitalita:
`std::isprint()` will assert (at least, on Windows) if the integer checked is < 0. Have it check the uint8_t value instead.

### DIFF
--- a/src/ELF/Binary.cpp
+++ b/src/ELF/Binary.cpp
@@ -536,7 +536,7 @@ Binary::string_list_t Binary::strings(size_t min_size) const {
   current.reserve(100);
 
   for (size_t i = 0; i < data.size(); ++i) {
-    char c = static_cast<char>(data[i]);
+    uint8_t c = data[i];
 
     // Terminator
     if (c == '\0') {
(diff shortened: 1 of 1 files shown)


romainthomas:
That's weird [...] Could you add the different steps to trigger the assertion?

nikitalita:
I just modified the elfreader example to dump the string table:

```c++
// add this to the bottom of elf_reader.cpp
  auto strings = binary->strings();
  if (strings.size() > 0) {
    std::cout << "== Strings ==" << '\n';
    for (const std::string& string : strings) {
      std::cout << string << '\n';
    }
  }
```

This was using Visual Studio 2022, Debug build.

romainthomas:
I tested this code (in Debug with the Debug runtime) and I can't reproduce the issue

Screenshot_2023-04-01_14-57-50

nikitalita:
here's the fail case: 
image

if you cast the a uint8_t with the value 250 (or indeed, any number from 129-255) to a char and attempt to use `std::isprint()`, it will assert. This is what we're doing in `strings()` currently.

romainthomas:
Ok got it. Thanks for the details.

PR #891, opened 2023-04-07
Fix missing <cstdint> header for gcc 13

topazus:
ref: https://gcc.gnu.org/gcc-13/porting_to.html#header-dep-changes


romainthomas:
Thank you!

PR #894, opened 2023-04-14
use correct .shstrtab section name when relocating it

kilograham:
the code `std::string sec_name = binary_->shstrtab_name();` was referencing the section right after it was deleted, leading to unexpected results (possibly returning ".shstrtab" by default, but also returning a different section's name if sections have been added).


romainthomas:
Nice catch! Thank you.

PR #900, opened 2023-04-19
[PE] fix has_relocations_ field

spobit:


src/PE/Binary.cpp
```diff
@@ -705,11 +705,13 @@ Binary::it_const_relocations Binary::relocations() const {
 Relocation& Binary::add_relocation(const Relocation& relocation) {
   auto newone = std::make_unique<Relocation>(relocation);
   relocations_.push_back(std::move(newone));
+  this->has_relocations_ = this->relocations_.size();
```
romainthomas, line 708: I would prefer `.empty()` overs `.size()`

src/PE/Binary.cpp
```diff
@@ -705,11 +705,13 @@ Binary::it_const_relocations Binary::relocations() const {
...
   return *relocations_.back();
 }
 
 void Binary::remove_all_relocations() {
   relocations_.clear();
+  this->has_relocations_ = this->relocations_.size();
```
romainthomas, line 714: Same

romainthomas (changes requested):
cf. comments

src/PE/Binary.cpp
spobit: good idea

romainthomas:
Merged through 4ef1759bd484c37289675cf4d05862d924435f9f (I squashed the commits)

PR #924, opened 2023-06-07
feat(macho): allow to create DynamicSymbolCommand in Python

LouisBrunner:
I created a static function (`createForPython`) to avoid messing with the implementation of the default constructor.

### DIFF
--- a/include/LIEF/MachO/DynamicSymbolCommand.hpp
+++ b/include/LIEF/MachO/DynamicSymbolCommand.hpp
@@ -48,8 +48,6 @@ class LIEF_API DynamicSymbolCommand : public LoadCommand {
 
   DynamicSymbolCommand(const details::dysymtab_command& cmd);
 
-  static DynamicSymbolCommand createForPython();
-
   DynamicSymbolCommand& operator=(const DynamicSymbolCommand& copy);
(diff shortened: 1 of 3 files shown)


include/LIEF/MachO/DynamicSymbolCommand.hpp
```diff
@@ -48,6 +48,8 @@ class LIEF_API DynamicSymbolCommand : public LoadCommand {
 
   DynamicSymbolCommand(const details::dysymtab_command& cmd);
 
+  static DynamicSymbolCommand createForPython();
```
romainthomas, line 51: Core header files should not expose Python specific function :)

romainthomas (changes requested):
I get the idea behinds this PR and I'll handle the integration in the next couple of days.

romainthomas:
Let me know if 1a69a1b16bf7f15cf92a83da0f3584446b5a6f65 would be ok.

romainthomas:
ping @LouisBrunner

LouisBrunner:
Hey, sorry for the delay.

It's not possible for me to test running an actual binary without all the other changes for `MainCommand`, etc, however from `otool`'s output it looks like it's working.

PR #947, opened 2023-08-06
CMake: Fix checking Windows platform

Biswa96:
In CMake, WIN32 is true for any toolchain targeting Windows platform.
    See https://cmake.org/cmake/help/latest/variable/WIN32.html


romainthomas:
Thank you!

PR #949, opened 2023-08-07
Removed _ITERATOR_DEBUG_LEVEL override on Windows builds

### LINKED ISSUE #948: Let CMake automagically pick `_ITERATOR_DEBUG_LEVEL` for Windows
**Describe the bug**
Windows doesn't seem to let you link two objects when there is a mismatch between the value of `_ITERATOR_DEBUG_LEVEL` between them. The default value for `_ITERATOR_DEBUG_VALUE` on a Debug build is `2`, but LIEF is manually overwriting that value with `0`, which makes the linker not allow the linkage between LIEF and a standard Debug binary.
Removing the following lines made everything work again: https://github.com/lief-project/LIEF/blob/master/CMakeLists.txt#L165-L168

**To Reproduce**
Statically compile LIEF (Debug) against a Debug shared library without deleting the lines mentioned (will throw `_ITERATOR_DEBUG_LEVEL mismatch` error)

**Expected behavior**
LIEF should allow CMake to pick the value for this compiler variable instead of overriding it. That would allow Debug binaries to compile and link normally.

 - System and Version : Windows 7 x64 - Visual Studio 2022 (Compiling for x86)
 - Target format (PE, ELF, Mach-O): PE (not relevant)
 - LIEF commit version: bab28ae0628fc6b6fa29fcafcfb7277f48ce0f40

**Additional context**
FIX: Deleting the following lines (165-168) from `CMakeLists.txt`:
```c
if(CMAKE_BUILD_TYPE MATCHES Debug AND WIN32)
  add_definitions(-D_ITERATOR_DEBUG_LEVEL=0 -D_SECURE_SCL=0
                  -D_HAS_ITERATOR_DEBUGGING=0)
endif()
```

rdbo:
Fixes #948 
Summary:
Windows doesn't seem to let you link two objects when there is a mismatch between the value of `_ITERATOR_DEBUG_LEVEL` between them. The default value for `_ITERATOR_DEBUG_VALUE` on a Debug build is `2`, but LIEF is manually overwriting that value with `0`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -162,11 +162,6 @@ if(MSVC)
   add_compile_options(/bigobj)
 endif()
 
-if(CMAKE_BUILD_TYPE MATCHES Debug AND WIN32)
-  add_definitions(-D_ITERATOR_DEBUG_LEVEL=0 -D_SECURE_SCL=0
-                  -D_HAS_ITERATOR_DEBUGGING=0)
-endif()
-
 if(CMAKE_BUILD_TYPE MATCHES Debug AND UNIX)
   target_compile_options(LIB_LIEF PRIVATE -g -O0)
 endif()


romainthomas:
Good to me!

PR #951, opened 2023-08-16
Fix dos_stub write in PE builder

alga81:
Dear LIEF team!
This PR fixes the write of the PE's dos_stub, corrupted after invoking `LIEF::PE::Builder::build()`.

Reproduce with any Windows PE located at `in_path`:
```
int main(int argc, char** argv)
{
    auto pe = LIEF::PE::Parser::parse(in_path);
    LIEF::PE::Builder pe_builder(*pe);
    pe_builder.build();
    pe_builder.write(out_path);
    return 0;
}
```

`out_path` is corrupted:

dos_stub_bug

After this change, the original dos_stub is correctly stored.

Thanks for this excellent library!
Alex


romainthomas:
Resolved through 660f26ebf499d499d08d442a77ebc8cc835fab72

PR #953, opened 2023-08-20
Use list comprehension instead of map/filter in tutorial

gvalkov:
It's slightly clearer than a `next(map(filter(lambda ...)))` (imho).


romainthomas:
Yes it makes sense! Thank you

PR #961, opened 2023-09-08
Adjust file listings for Ninja Multi-Config build

starseeker:
For CMake's Ninja Multi-Config generator, we need to list all the mbedtls files explicitly as BUILD_BYPRODUCTS for ExternalProject_Add rather than just the directory.


romainthomas:
Great! Thank you!

Issue #796, opened 2022-09-17
MIPS ELF imported symbols are not recognized correctly


martonilles:
**Describe the bug**
I have the following file (same as in #795).
```
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

const int NUMBER = 42;
const char VERSION[] = "11.22";
const char *VERSION2 = "22.33";


void function1 (const char* str) {
   char buffer[16];
   strcpy(buffer, str);
}

int main(int argc, char *argv[]) {
   printf("Hello %s %s\n", VERSION, VERSION2);

   exit(0);
}
```
Which is compiled to MIPS platforms:
```
test-mips-64bit-le: test.c
        mips64el-linux-gnuabi64-gcc-10 -o $@ $^

test-mips-64bit-le-lib: test.c
        mips64el-linux-gnuabi64-gcc-10 -o $@ -shared $^

test-mips-64bit-be: test.c
        mips64-linux-gnuabi64-gcc-10 -o $@ $^
```

[test-mips.zip](https://github.com/lief-project/LIEF/files/9592653/test-mips.zip)

Except the shared lib (test-mips-64bit-le-lib), these parse correctly, but selecting imported symbols/function is failing.

**To Reproduce**
I have the following python code to show list imported symbols/functions:
```
import lief
import sys

lief.logging.set_level(lief.logging.LOGGING_LEVEL.DEBUG)
l = lief.parse(sys.argv[1])


print("SYMBOLS", [s.name for s in l.symbols])
print("FUNCTIONS", [s.name for s in l.functions])
print("IMPORTED SYMBOLS", [s.name for s in l.imported_symbols])
print("IMPORTED FUNCTIONS", [s.name for s in l.imported_functions])
```

Running it results in the following:
```
[...]

PR #968, opened 2023-09-19
Remove verification of the non st_value for import function.

Mrbenoit624:
mips and powerpc add a st_value for their import function (issue #796). So we need to remove this verification to let not to hide some import functions

### DIFF
--- a/src/ELF/Symbol.cpp
+++ b/src/ELF/Symbol.cpp
@@ -214,8 +214,9 @@ bool Symbol::is_imported() const {
   // An import must not be defined in a section
   bool is_imported = shndx() == static_cast<uint16_t>(SYMBOL_SECTION_INDEX::SHN_UNDEF);
 
-  // An import must not have an address
-  is_imported = is_imported && value() == 0;
+  // An import must not have an address but mips and powerpc put an addres for
+  // import function.
+  //is_imported = is_imported && value() == 0;
 
   // its name must not be empty
   is_imported = is_imported && !name().empty();


src/ELF/Symbol.cpp
```diff
@@ -214,8 +214,9 @@ bool Symbol::is_imported() const {
   // An import must not be defined in a section
   bool is_imported = shndx() == static_cast<uint16_t>(SYMBOL_SECTION_INDEX::SHN_UNDEF);
 
-  // An import must not have an address
-  is_imported = is_imported && value() == 0;
+  // An import must not have an address but mips and powerpc put an addres for
```
romainthomas, line 217: Would you have an example/sample for this?

Mrbenoit624: Yes of course, https://transfer.ci-yow.com/d?id=RcdhAdQ3vsAuWeX, this is one the function imported "strstr" which have an st_value.

I have some better example if you want but I can't share you here.

romainthomas: No thanks it's a good sample.
I see the issue, my only concern is that your change drop the condition for **all** the architectures, not only MIPS/PPC. 
On the other hand, the targeted architecture is not accessible from the `Symbol` class. 
I need to think about how we can fix this design issue.

Mrbenoit624: I am agree, I would like to limited this drop this check only for these two architectures but as you said, I didn't find how to do it without a big refactorisation of you code...

But when I check how some technique to identify imported function, and how the loader do, it should look symbols and resolve *undefined symbols*. So the main meaning check should be:
```
  bool is_imported = shndx() == static_cast<uint16_t>(SYMBOL_SECTION_INDEX::SHN_UNDEF);
```
the others is for safety but they should not add bug.

I let you see what you prefer and if you would like to do a big factorisation to get access of ARCH by symbols, I could help you to dev :).

PR #971, opened 2023-09-21
ELF Python - fixing utf-8 encoding issue

su-vikas:
While trying to get the strings from an ELF file, I was getting the following error: 

`Exception in print_strings: 'utf-8' codec can't decode byte 0xaa in position 0: invalid start byte`

This PR contains the fix for the issue, and derived from this [previous commit](https://github.com/lief-project/LIEF/commit/c13d0c54fee914adf6b0fabc99801fd517e2d559)

I am observing this issue with all the ELF files I have tried. This can be reproduced by using the elf_reader.py example with `--strings` option. 

 - System and Version : MacOS 13.5.1, Python 3.11.5
 - Target format: ELF 
 - LIEF commit version:  Using the latest code from master.

### DIFF
--- a/api/python/src/ELF/objects/pyBinary.cpp
+++ b/api/python/src/ELF/objects/pyBinary.cpp
@@ -45,6 +45,7 @@
 
 #include "pyIterator.hpp"
 #include "pyErr.hpp"
+#include "pySafeString.hpp"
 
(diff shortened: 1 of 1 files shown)


romainthomas:
All good, thank you!

PR #978, opened 2023-10-09
Prefer CMAKE_CURRENT_BINARY_DIR over CMAKE_BINARY_DIR

adrianherrera:
Using `CMAKE_BINARY_DIR` breaks nesting LIEF within another cmake project via add_subdirectory (as described [here](https://lief-project.github.io/doc/latest/installation.html#add-subdirectory-or-fetchcontent)). This is because `CMAKE_BINARY_DIR` will return the top-level of the build tree, but if LIEF is nested within another cmake project this will not correspond to LIEF's actual location.

## Steps to reproduce

```bash
git clone https://github.com/lief-project/LIEF
cd LIEF/examples/cmake/add_subdirectory
git clone https://github.com/lief-project/LIEF

mkdir build
cd build
cmake ..
make
```

You should see an error like this:

```
LIEF/examples/cmake/add_subdirectory/LIEF/src/hash_stream.cpp:18:10: fatal error: mbedtls/md.h: No such file or directory
```

## Fix

Instead, `CMAKE_CURRENT_BINARY_DIR` will use the correct location.


romainthomas:
Thanks @adrianherrera. It sounds good to me!

@starseeker this line was introduced by your in #961. It's still compatible with Ninja multi-config?

starseeker:
Tested and yes, it looks like this change is fine with Ninja Multi-Config generator.

adrianherrera:
Thanks folks! 🙌

PR #1011, opened 2024-01-11
Fix the size of bignum_wrapper_t returned from LIEF::PE::RsaInfo methods

ch4mchi:
When initializing `bignum_wrapper_t`, returned by `RsaInfo::N()`, `RsaInfo::E()`, `RsaInfo::D()`, `RsaInfo::P()`, `RsaInfo::Q()`, it is initialized with the bit size instead of the byte size, leading to the problem of zeros being filled at the beginning of the actual content.

Since `bignum_wrapper_t` is a `std::vector<uint8_t>`, it should be initialized with the byte size, not the bit size.

This commit fixes the issue by using `mbedtls_mpi_size` instead of `mbedtls_mpi_bitlen` during initialization.


romainthomas:
Indeed. Thank you!

PR #1025, opened 2024-02-23
Add usable constructor for MachO::BuildVersion

jevinskie:
I use this in [dylibify](https://github.com/jevinskie/dylibify/blob/7c84ace9b0efb6d458f243575d60b0d7473130d6/cpp/dylibify-lief-cpp.cpp#L286-L300) to modify (via delete then create [what this PR is for]) the platform from iOS to macOS while converting the executable to a dylib. Similar type functionality is provided by vtool in Xcode/Command Line Tools.


romainthomas:
Sounds good. Thank you!

PR #1029, opened 2024-02-28
fix the bug in symbol.section_number to section

wonderkun:
According to the coff format document : https://www.delorie.com/djgpp/doc/coff/symtab.html

The number of the section that this symbol belongs to. The first section in [the section table](https://www.delorie.com/djgpp/doc/coff/scnhdr.html) is section one.

image


romainthomas:
Looks good! Thank you!

PR #1047, opened 2024-04-23
fix ASN1Reader::tag2str

### LINKED ISSUE #1046: `lief.PE.parse' function throws an out-of-range exception
[test.zip](https://github.com/lief-project/LIEF/files/15071882/test.zip)
**Describe the bug**
The `lief.PE.parse' function throws an out-of-range exception for some signature-corrupted PE binaries.

**To Reproduce**
1. Build LIEF and the `pe_reader` example.
2. Run `pe_reader` for the (unzipped) attached sample.
3. You will then see the following output and the exception will be thrown:
```bash
PE Reader
padding: 0
Checksum               : 0x00f823
OptionalHeader.checksum: 0x011f4c
DOS stub: @0x40:0xc0
Parsing rich header
Offset to rich header: 0xa0
XOR key: 0xed389f2d
ID:           0x0102
Build Number: 0x6c36
Count:        0x1
ID:           0x00ff
Build Number: 0x6c36
Count:        0x1
ID:           0x0109
Build Number: 0x6c36
Count:        0x1
ID:           0x0001
Build Number: 0x0000
Count:        0x65
ID:           0x0101
Build Number: 0x6665
Count:        0x2
ID:           0x0101
Build Number: 0x6b81
Count:        0x5
ID:           0x0103
Build Number: 0x6b81
Count:        0x2
ID:           0x0104
Build Number: 0x6b81
Count:        0x10
ID:           0x0105
Build Number: 0x6b81
Count:        0x17
ID:           0x0093
Build Number: 0x7809
Count:        0x10
Parsing sections
Number of sections that could be added: #5
Processing Import Table
Parsing signature
Signature Offset: 0x3000
Signature Size:   0x1c18
Signature 32r0x1 (0x20 bytes)
terminate called after throwing an instance of 'std::out_of_range'
  what():  basic_string::substr: __pos (which is 3) > this->size() (which is 0)
Aborted
```
**Expected behavior**
`pe_reader` does not throw the `std::out_of_range` exception.

System and Version : Ubuntu 22.04 on WSL2 on Windows 10
Target format: PE (AMD64)
LIEF commit version: 0.15.0-573c885d

**Additional context**
The cause seems to be that the `substr` function in the `std::string ASN1Reader::tag2str` function does not take into account the case where the string may be an empty string.
So I suggest adding a string length check before executing `substr`.

I will make the pull request to fix this.

The attached sample is not a malware, but was created (by me) with reference to the structure of the following sample:
https://www.virustotal.com/gui/file/c90b935c9ee9f0ef186fd170055ae6ad32a2350d24f3d7c766b41066fe2d6cba

**NOTICE**

If the issue does not contain enough information to be reproduced,
it will be flagged as [incomplete](https://github.com/lief-project/LIEF/issues?q=is%3Aissue+is%3Aclosed+label%3AIncomplete)
and closed.

**/NOTICE**

yuki-mogi-ffri:
This PR will fix #1046.

This pull request adds a string length check before executing the `substr` in the `std::string ASN1Reader::tag2str` function.

Specifically, the function has been changed to return `MBEDTLS_ASN1_UNKNOWN` if the size of `tag_str` is less than 3.

### DIFF
--- a/src/BinaryStream/ASN1Reader.cpp
+++ b/src/BinaryStream/ASN1Reader.cpp
@@ -473,6 +473,10 @@ std::string ASN1Reader::tag2str(int tag) {
   HANDLE(ASN1_CONSTRUCTED);
   HANDLE(ASN1_CONTEXT_SPECIFIC);
 
+  if (tag_str.size() < 3) {
+    return "MBEDTLS_ASN1_UNKNOWN";
+  }
+
   return tag_str.substr(3);
 #undef HANDLE
 }


romainthomas:
Look good! Thank you.

PR #1053, opened 2024-05-03
Rename DEBUG enum variants

lelandjansen:
Many C++ build systems define `DEBUG=1` to signify when they're building in debug mode. Variables, enum variants, etc. named `DEBUG` are subject to this macro substitution, get renamed to `1`, and fail to compile. This PR renames `DEBUG` enum variants to a similar name that doesn't get substituted by the preprocessor.

**Test plan**

```bash
mkdir build
cd build
cmake -DCMAKE_BUILD_TYPE=Release ..
cmake --build . --target LIB_LIEF --config Release
```


romainthomas:
Looks good, I'll take a closer CI-review in two days!

romainthomas:
@lelandjansen thank you for your PR. I did some update on the Python side and I merged through ef824c71bc6c58a5728487cafc2abac0bac9b4cf

lelandjansen:
Thank you @romainthomas!

PR #1067, opened 2024-06-05
Add additional missing undef for ELF

husixu1:
Program fails to build when `elf.h` (or `boost/dll/shared_library.hpp`) is included before `LIEF.hpp`.
Example: 
```cpp
// test.cpp, g++ -std=c++17 test.cpp
#include <elf.h>
#include "LIEF/LIEF.hpp"
int main(void) {
    return 0;
}
```
Environment: 
LIEF: v0.14.1
OS: Arch Linux x86_64 6.9.3-arch1-1
GCC: 14.1.1

### DIFF
--- a/include/LIEF/ELF/undef.h
+++ b/include/LIEF/ELF/undef.h
@@ -484,6 +484,7 @@
 
 #undef PT_GNU_STACK
 #undef PT_GNU_RELRO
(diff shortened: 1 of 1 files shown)


romainthomas:
I refactored the enums in `main` so this workaround is no longer needed :)

PR #1084, opened 2024-07-30
Fix some more typos

pmeerw:


romainthomas:
Thanks @pmeerw for both PR  :+1:

PR #1096, opened 2024-08-30
Remove unnecessary imports that mess up type hints

### LINKED ISSUE #1095: Type hints aren't working properly
Currently, many of the type hint files (`.pyi`) import classes that they hint from the library.
For example, `lief\__init__.pyi` contains the lines:

```python3
import lief.Binary # type: ignore
import lief.Function # type: ignore
import lief.DebugInfo # type: ignore
```

However, the file provides hints for those two classes!

```python3
class Binary(Object):
    ...

class Function(Symbol)
    ...

class DebugInfo:
    ...
```

This confuses `mypy`, so I get no hints for those types.

This behavior makes sense - a `pyi` file "shadows" a python file for the duration of the type check, so you are importing... from the file itself...?

The correct solution would be to not import classes that are hinted within the `pyi` file.

omer54463:
Resolves #1095

### DIFF
--- a/api/python/lief/DEX.pyi
+++ b/api/python/lief/DEX.pyi
@@ -3,12 +3,6 @@ from typing import Any, ClassVar, Optional, Union
 from typing import overload
 import io
 import lief.DEX # type: ignore
-import lief.DEX.Class # type: ignore
-import lief.DEX.File # type: ignore
-import lief.DEX.MapItem # type: ignore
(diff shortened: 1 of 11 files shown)


romainthomas:
Closing (c.f. https://github.com/lief-project/LIEF/issues/1095#issuecomment-2322892680)

PR #1100, opened 2024-09-08
Fix security_cookie type

ShallowFeather:
Hi, I found that the return type of security_cookie is wrong, just change it's type from `uint32_t` to `uint64_t`


romainthomas:
Thank you for this fix!

PR #1105, opened 2024-09-16
Add rustls-tls feature to reqwest

Huntragon:
The rust bindings use reqwest to download prebuilt binaries and link against those. This is done via an HTTPS request, and TLS is provided by using openssl. When building via nix, this is an issue since it requires having openssl present so it can link against it.
By adding the rusttls-tls feature to reqwest, it uses a pure rust implementation of TLS, simplifying the building process.


romainthomas:
Thanks @Huntragon 

It makes sense to avoid requiring openssl if there is a pure Rust TLS implementation.

Thank you for this improvement

PR #1106, opened 2024-09-17
Fix bug in Python PE Relocation bindings caused by copy

mikenawrocki:
Python: mark add_entry return value as reference
Makes the return value from the PE::Reference add_entry method an internal reference.

If this is not the case, a copy occurs which sets the new object's `relocation_` field to null (https://github.com/lief-project/LIEF/blob/main/src/PE/RelocationEntry.cpp#L36)

In addition/alternately, should the copy constructor also copy `other.relocation_` instead of assigning to null?

To replicate the bug:
```
import lief
r = lief.PE.Relocation()
r.virtual_address = 0x1000

re = lief.PE.RelocationEntry()
re.data = 0x3123

re2 = r.add_entry(re)

print(f"Addr: {hex(re2.address)}")
```
should print 0x1123, but instead prints 0x123


romainthomas:
Nice catch!

PR #1111, opened 2024-09-24
Use Attribute.TYPE instead of SIG.ATTRIBUTE_TYPES in authenticode example

TTMaZa:
Hi there.

Inspired by this change: https://github.com/lief-project/LIEF/commit/8dd8a255a824f4f266d238133f8a8b2942198f5b

I got the impression, that this should be changed as well.

Not sure if there is anything else to do.

Let me know if I can be of any help with this.


romainthomas:
Hi @TTMaZa

You are right. Thank you

PR #1116, opened 2024-10-15
Speed up ELF symbol lookup

wannacu:
It's too slow to rebuild an ELF object that contains a large number of symbols. This happens because the frequent linear looking up was expensive.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ dist/
 api/rust/cargo/target/*
 api/rust/examples/target/*
 Cargo.lock
+.cache
--- a/examples/cpp/CMakeLists.txt
+++ b/examples/cpp/CMakeLists.txt
@@ -75,7 +75,7 @@ foreach(example ${LIEF_CPP_EXAMPLES})
   set_target_properties(
     ${output_name}
     PROPERTIES POSITION_INDEPENDENT_CODE ON
-               CXX_STANDARD              11
+               CXX_STANDARD              17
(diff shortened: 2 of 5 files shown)


romainthomas:
Hi @wannacu 

Thank you for this interesting PR!.
Some remarks though:

1. It uses `std::string_view` in public headers which raises the minimum C++ version from 11 to 17. I would like to stick to C++11 for the API
2. What if a user add a symbol? From your implementation it seems that the invariant is no longer valid isn't?

peledins-zimperium:
This is a very important fix and would enable removing symbols en-mass from LIEF.
However, there is a corner case not handled: there can be 2 symbols that go by same name, for example, when same static function is defined in 2 source files, e.g.:
```
objdump --syms ./a.out 

./a.out:	file format mach-o arm64

SYMBOL TABLE:
0000000100003f70 l     F __TEXT,__text __ZL3barv
0000000100003f98 l     F __TEXT,__text __ZL3barv
0000000100003f84 g     F __TEXT,__text __Z3foov
0000000100000000 g     F __TEXT,__text __mh_execute_header
0000000100003f50 g     F __TEXT,__text _main
0000000100004000 g     O __DATA,__common _z
``` 

bar() is static.

k4lizen:
> there can be 2 symbols that go by same name

this should be easily fixable with `std::multimap` right?
is this and not using std::string the only blocker for this PR?

PR #1132, opened 2024-12-10
Update segment offsets in MachO::Binary::extend_segment

DzenIsRich:
`LIEF::MachO::Binary::extend_segment` changes `file_offset` of some segments, therefore we need to call `refresh_seg_offset` to update offset cache.


romainthomas:
Cherry picked on `release/0.16.x`

PR #1133, opened 2024-12-10
MachO::Binary::shift_command: fix exports of dyld

DzenIsRich:
Despite the field name `address` it stores a file offset.

### DIFF
--- a/src/MachO/Binary.cpp
+++ b/src/MachO/Binary.cpp
@@ -816,7 +816,7 @@ void Binary::shift_command(size_t width, uint64_t from_offset) {
 
   if (DyldExportsTrie* exports = dyld_exports_trie()) {
     for (ExportInfo& info : exports->exports()) {
-      if (virtual_address >= info.address()) {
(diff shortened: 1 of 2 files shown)


romainthomas:
A binary sample or a test case that triggers this issue is welcomed (to avoid regression)

DzenIsRich:
I cannot reproduce this problem on `main` branch, it seems to me builder or some other component fixes incorrect exports somewhere else.

I tried the following binary:
```cpp
int weak_sym() __attribute__((weak));
int weak_sym() { return 0; }

int main() {
  if (weak_sym)
    return weak_sym();
  else
    return -1;
}
```

```bash
clang -O3 weak-sym.c -Wl,-no_fixup_chains -o MachO64_AArch64_weak-sym.bin
```

And use it in `test_symbols.py::test_symbol_shift`.

romainthomas:
Ok thank you. I'll add them to the sample archive and merge the PR

PR #1134, opened 2024-12-10
MachO::Binary: treat all sections in the same way during shift

DzenIsRich:


romainthomas:
I welcome a test case that ensures we don't regress this fix.

DzenIsRich:
It is not trivial to craft a test case that reproduces the problem on main branch. I can reproduce the problem with proprietary binary that I cannot share, but it's difficult to minimize it.

Can we proceed without tests?

romainthomas:
Ok I trust you on this. I'll see if I can find a testcase on my end.

PR #1135, opened 2024-12-10
MachO::Binary::remove(Symbol): search symbol by ptr instead of name

DzenIsRich:
The previous implementation was searching symbol by name, due to this it could delete other symbol with the same name, instead of `sym` that was passed to the function.


romainthomas:
Cherry picked on the `release/0.16.x` branch

PR #1136, opened 2024-12-10
MachO::Binary: improve code related to shift method

DzenIsRich:
This commit outlines common pattern in the code into `ensure_command_space` method that checks if there is enough space between the command table and the first section.

Additionally this commit changes `add_section` alignment from a size of a page to the section alignment requested by user.

### DIFF
--- a/tests/macho/test_builder.py
+++ b/tests/macho/test_builder.py
@@ -13,6 +13,10 @@ from utils import get_sample, is_apple_m1, is_osx, is_x86_64, sign, chmod_exe, i
 
 lief.logging.set_level(lief.logging.LEVEL.INFO)
 
+def align_to(value, alignment):
+    # llvm::alignTo
+    return (value + alignment - 1) & ~(alignment - 1)
+
 def dyld_check(path: str):
     dyld_info_path = "/usr/bin/dyld_info"
     if not pathlib.Path(dyld_info_path).exists():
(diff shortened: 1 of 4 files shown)


DzenIsRich:
I've added a test only for functional change: that `add_section` prepends sections continuously without making gaps.

I think this PR is ready for review.

tests/macho/test_builder.py
```diff
@@ -164,6 +168,24 @@ def test_add_section_id(tmp_path):
...
+            continue
+        sections = sorted([s for s in segment.sections if s.type not in bad_section_types], key=lambda s: s.offset)
+        next_expected_offset = sections[0].offset
+        print(segment)
+        for section in sections:
+            print(section)
+            next_expected_offset = align_to(next_expected_offset, 1 << section.alignment)
+            assert section.offset == next_expected_offset
+            next_expected_offset += section.size
+
```
DzenIsRich, line 188: I think I could move this code into `lief.MachO.check_layout`, but I am afraid of breaking lots of other tests/assumptions; this could be a follow-up.

romainthomas: I currently working on updating the `check_layout` function to include latest checks from Apple. I'll add this check as well

tests/macho/test_builder.py
```diff
@@ -13,6 +13,10 @@
 
 lief.logging.set_level(lief.logging.LEVEL.INFO)
 
+def align_to(value, alignment):
+    # llvm::alignTo
+    return (value + alignment - 1) & ~(alignment - 1)
```
romainthomas, line 18: Feel free to backport this implementation in the `utils.hpp` as well which is absolutely not optimized

DzenIsRich: I did consider this, but in the end I decided that I don't want to do it.
The current implementation in `utils.hpp` allows alignment to any number, however this code from LLVM supports alignment only for powers of two.
That's why `utils.hpp` should be changed separately and carefully, ensuring that none uses this function with an argument that is not power of two.

And I just realized I need to add an assert that `alignment` is power of two here.

DzenIsRich:
Please let me know if you expect me to do anything else in this PR (e.g. code changes, or add another test)

tests/macho/test_builder.py
romainthomas: Make sense!

PR #1137, opened 2024-12-10
MachO::binary: add extend_section

DzenIsRich:
This PR depends on #1136


DzenIsRich:
This turned out to be more difficult than I anticipated, especially after you've added layout_check, and I created a test that requires passing this new layout_check.

@romainthomas, I think, this PR is ready for review.

It would be nice to document somewhere down, that if user adds and extends sections with unusual alignment requirements in some _weird_ order, then it might cause errors in layout_check.
For example, user allocates two empty sections with alignment 1 and 256.
```
_text: 0x7fc -> 0x800
_a1:   0x7fc -> 0x7fc (empty for now)
_a256: 0x7fc -> 0x7fc (empty for now)
```

Then user extends both sections `_a1` and `_a256` to 1 (mind the order).
```
_text: 0x7fc -> 0x800
_a1:   0x7fb -> 0x7fc (size=1)
_a256: 0x7fb -> 0x7fb (empty for now)
```
```
_text: 0x7fc -> 0x800
_a1:   0x7fb -> 0x7fc (size=1)
_a256: 0x600 -> 0x601 (size=1)
```

Why `_a256` is at offset 0x600? Remember, we allocate everything towards the negative addresses, therefore the closest offset to `0x7fb` that satisfies alignment of `0x100` is `0x600`.

And we want to allocate everything in front of the `__text` section, because there might be metadata in `__DATA` segment that has unobvious relative offsets to `__text` section, therefore we want to keep the distance between `__text::end` and `__DATA::start` unchanged.

DzenIsRich:
And I just realized that we must keep track of alignment for _virtual addresses_. But let's fix that in a follow-up, this PR alone is big enough already.

DzenIsRich:
> But let's fix that in a follow-up, this PR alone is big enough already.

No, I cannot leave it like that, because we would produce binaries which may crash due to unaligned access.

peledins-zimperium:
I think this can be improved a bit - let's discuss f2f, let's not merge it in yet.

DzenIsRich:
> I think this can be improved a bit - let's discuss f2f @DzenIsRich , let's not merge it in yet.

We had a discussion, @peledins-zimperium has suggested a better implementation that ensures we move sections together, such that we preserve relative positions of sections from the left and from the right relatively to point that is shifted (obviously we cannot maintain relativeness property across the point of shift).

Long story short, the suggestion is to calculate maximum common alignment, and shift all sections by this value, instead of accumulating shift value.

romainthomas:
Ok thank you for the details. Since this functionality was not original present in LIEF, I trust you about the implementation.

DzenIsRich:
@peledins-zimperium, I've applied your suggestions, please review.

DzenIsRich:
During the review we discovered a bug described in #1144. As it's an extreme case with alignment values, I've opened a follow-up issue.

PR #1153, opened 2025-01-07
add support for spc link / urls in spc indirect data

bhoehl:
@romainthomas 
My use case of LIEF is reading signatures and validating them.
This is not necessarily using executables, instead I am validating fonts.
The Spc Indirect Data can contain other values instead of files, also urls.
This PR adresses the issue to support urls too.
If possible please review and merge, as I would really like to continue using this lib.
There is no real support for PKCS#9 for python out there ...

Thanks
Bernd

### DIFF
--- a/api/rust/cargo/lief-ffigen/Cargo.toml
+++ b/api/rust/cargo/lief-ffigen/Cargo.toml
@@ -8,7 +8,7 @@ license.workspace = true
 
 [dependencies]
 cc            = "1.0"
-autocxx-build = "0.27.0"
+autocxx-build = "=0.27.0"
 cmake         = "0.1.50"
 toml          = "0.8.6"
 serde         = "1.0"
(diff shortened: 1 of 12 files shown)


romainthomas:
Interesting use case :) Would you mind sharing a sample of a font using this PKCS #9 signature?

bhoehl:
@romainthomas 
an example of a such a signature
I am loading this file using:

LIEF::logging::set_level(LIEF::logging::LEVEL::DEBUG);
auto sig = SignatureParser::parse(data, false);

[signature.zip](https://github.com/user-attachments/files/18394990/signature.zip)

romainthomas:
Sorry for the delay. I merged the PR with additional tests in this commit: ac3b57deb100c361525573ee5100f6fba7149e6b

bhoehl:
@romainthomas I wanted to use that code in the python library, but it is not out yet as it seems.
is there any chance to get this as a wheel file ?

thanks 
~bernd

PR #1154, opened 2025-01-07
MachO::Binary::extend_section fill gap by default

DzenIsRich:
The previous implementation extended section by strictly requested amount, possibly creating a gap in case we needed to satisfy alignment requirements.

However, information about the size of the gap is lost, and not trivial to recover. This patch changes `extend_section` to extend section by more than requested to avoid creating the gap.


romainthomas:
Merged through: 9a4f609aa675571d506a54b5d97ebcf767cb338a

PR #1171, opened 2025-02-13
Dex: add getter for registers count

willboka:
- Add a getter for nb_registers from CodeInfo and update Python bindings as it is useful when guessing which registers are used for parameters.
- Add test for it using a known method from the provided Dex sample.
- Separate test_kik_methods() using both externally (default) and an internally defined method from the sample. I need to do that because it would not be reliable enough to test only with and externally defined method.


include/LIEF/DEX/Method.hpp
```diff
@@ -86,6 +86,8 @@ class LIEF_API Method : public Object {
   /// ACCESS_FLAGS as an std::set
   access_flags_list_t access_flags() const;
 
+  // CodeInfo to get additional data about a method i.e. argument count
+  const CodeInfo code_info() const;
```
romainthomas, line 90: Should return by const ref to avoid a copy :)

romainthomas:
Looks good! Just a minor comment

include/LIEF/DEX/Method.hpp
willboka: Yes, I fixed it in [3258b01](https://github.com/lief-project/LIEF/pull/1171/commits/3258b01209ee1c760a54be3a38c03d3e3fd256ae)

willboka: Yes, I fixed in d579053d543e2b29b030e5f4b7dabfeb303278f3

PR #1184, opened 2025-04-02
Add cstdio include

kraj:
This is needed for uintXX_t types, GCC-15 is defauling to C23 and will error about it.

e.g.

include/LIEF/PE/LoadConfigurations/VolatileMetadata.hpp:145:3: note: 'uint32_t' is defined in header '<cstdint>'; this is probably fixable by adding '#include <cstdint>'


romainthomas:
Thank you!

PR #1196, opened 2025-05-01
Use the furthest section when calculating the new file offset for the phdr table

u1f98e:
Currently relocate_phdr_table_v3() uses the furthest loadable segment's file offset when calculating a new file offset for the table, which causes it to sometimes relocate on top of non-loadable sections (like debug info), corrupting them.

### DIFF
--- a/src/ELF/Binary.cpp
+++ b/src/ELF/Binary.cpp
@@ -2649,9 +2649,9 @@ uint64_t Binary::relocate_phdr_table_v3() {
   static constexpr size_t USER_SEGMENTS = 10;
 
   uint64_t last_off = 0;
-  for (const std::unique_ptr<Segment>& segment : segments_) {
-    if (segment != nullptr && segment->is_load()) {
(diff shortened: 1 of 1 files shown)


romainthomas:
Hi @u1f98e 
Thank you for raising this point. Would you have an example/PoC that triggers this issue?

u1f98e:
While building a minimal example I realized the regions don't actually overlap (the sections past the last loadable one do get pushed back correctly), but the debuginfo does get corrupted after the shifting happens. `readelf` outputs this warning at the end of output, and gdb fails to load debug info:

```
readelf: Warning: Corrupt unit length (got 0x20f0500 expected at most 0xb26a) in section .debug_info
```

(For context this is a 32-bit RISC-V ELF, I haven't reproduced this with an x86 ELF yet).
This patch avoids shifting the debug sections, but it'd probably be better to fix the underlying issue. When I have more time I'll open a separate issue with better context, sorry.

PR #1198, opened 2025-05-02
Fix BinaryStream: Implement robust bounds check and fix warnings

kovalev0:
Replaces the bounds check in BinaryStream::peek_array and read_array with a version robust against unsigned
overflow when calculating pos_ + N.

The check '(pos_ + N) <= size()' is replaced by 'N <= size() - pos_' (combined with 'pos_ <= size()'). This form is safe
against unsigned overflow.

This change also removes redundant checks that caused -Wtype-limits warnings (tested on i586), as the robust
unsigned check supersedes them.

/usr/src/RPM/BUILD/liblief-0.17.0/include/LIEF/BinaryStream/BinaryStream.hpp: 195:38:
warning: comparison is always true due to limited range of data type [-Wtype-limits]
 195 |       && (static_cast<int64_t>(pos_) >= 0 && static_cast<int64_t>(N) >= 0)
     |           ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~
/usr/src/RPM/BUILD/liblief-0.17.0/include/LIEF/BinaryStream/BinaryStream.hpp:
196:42: warning: comparison is always true due to limited range of data type
[-Wtype-limits]
 196 |       && (static_cast<int64_t>(pos_ + N) >= 0);
     |          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~

### DIFF
--- a/include/LIEF/BinaryStream/BinaryStream.hpp
+++ b/include/LIEF/BinaryStream/BinaryStream.hpp
@@ -190,10 +190,7 @@ class LIEF_API BinaryStream {
     }
     // Even though offset + size < ... => offset < ...
     // the addition could overflow so it's worth checking both
-    const bool read_ok = pos_ <= size() && (pos_ + N) <= size()
-      /* Check for an overflow */
-      && (static_cast<int64_t>(pos_) >= 0 && static_cast<int64_t>(N) >= 0)
-      && (static_cast<int64_t>(pos_ + N) >= 0);
+    const bool read_ok = pos_ <= size() && N <= size() - pos_;
 
     if (!read_ok) {
       return make_error_code(lief_errors::read_error);


romainthomas:
Sorry I can't accept this PR. Your updates weaken the checks.

kovalev0:
> Sorry I can't accept this PR. Your updates weaken the checks.

You are likely referring to the check ensuring the end position (`pos_ + N`) does not numerically exceed `INT64_MAX`, which the original code implicitly performed through the `static_cast<int64_t>(...) >= 0` comparison ?
In that case I would suggest the following patch:

```patch
--- a/include/LIEF/BinaryStream/BinaryStream.hpp
+++ b/include/LIEF/BinaryStream/BinaryStream.hpp
@@ -21,6 +21,7 @@
 #include <cstring>
 #include <string>
 #include <algorithm>
+#include <limits>
 
 #include "LIEF/endianness_support.hpp"
 #include "LIEF/errors.hpp"
@@ -190,10 +191,8 @@ class LIEF_API BinaryStream {
     }
     // Even though offset + size < ... => offset < ...
     // the addition could overflow so it's worth checking both
-    const bool read_ok = pos_ <= size() && (pos_ + N) <= size()
-      /* Check for an overflow */
-      && (static_cast<int64_t>(pos_) >= 0 && static_cast<int64_t>(N) >= 0)
-      && (static_cast<int64_t>(pos_ + N) >= 0);
+    const bool read_ok = pos_ <= size() && N <= (size() - pos_)
+      && (pos_ + N) <= static_cast<uint64_t>(std::numeric_limits<int64_t>::max());
 
     if (!read_ok) {
       return make_error_code(lief_errors::read_error);
-- 
2.42.2
```
This version also does not produce a compile-time warning.

romainthomas:
Which version of gcc/clang triggers this warning?

kovalev0:
> Which version of gcc/clang triggers this warning?

```
$ arch 
i686
$ c++ --version
i586-alt-linux-g++ (GCC) 14.2.1 20241028 (ALT Sisyphus 14.2.1-alt1)
Copyright (C) 2024 Free Software Foundation, Inc.
```
The warning is triggered only when building on the i586 architecture; there is no such warning on x86_64 or aarch64.

kovalev0:
> $ arch 
> i686
> $ c++ --version
> i586-alt-linux-g++ (GCC) 14.2.1 20241028 (ALT Sisyphus 14.2.1-alt1)
> Copyright (C) 2024 Free Software Foundation, Inc.

Another interesting thing for this platform was that a problem was discovered using the ALT rpm verify-elf brp script:
`verify-elf: WARNING: ./usr/lib/libLIEF.so: uses non-LFS functions: fcntl fopen readdir stat`

As a workaround, I currently use a build with options:
`-D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64`

but it looks like this also requires a fix in the code?

PR #1216, opened 2025-05-23
Fixed issue with explicit constructor compilation when old (GCC 9.0) compiler used

ievgendeveloper:
The issue occurs when using older compiler versions (for example, **gcc/g++ 9.0** in my case). I’m forced to stick with this version because it supports **C++17**, and on **Ubuntu 16.04** (the **oldest Ubuntu LTS** release) no newer compiler is available in the repositories.

The crux of the problem is that result<T> is an alias:

```
template<typename T>
using result = tl::expected<T, lief_errors>;
```

tl::expected requires an explicit constructor for the type T. In other words, we need to return a result<T> that’s constructed in place via T’s explicit constructor—without any copies or moves. The following construct:

```
return result<FileStream>(tl::in_place, std::move(ifs), size);
```

solves the problem.


romainthomas:
LGTM

Issue #1219, opened 2025-05-31
Certificate OID parsing issues: failure to receive some OID strings and failure to decode non-roman characters


Squiblydoo:
**Describe the bug**
When parsing a code-signing certificate, LIEF fails to convert some OID to strings. LIEF also fails to decode non-roman characters in the certificate subject.

The failure to convert some OID may be related to the failure to decode the non-roman characters, but I'm not certain.

**To Reproduce**
Attempting to print the subject from the certificate from this file: [be5d6c4aa4b27548a06c2afaef3b4035abf65566e9a8bfd642b4a2032729656e](https://bazaar.abuse.ch/sample/be5d6c4aa4b27548a06c2afaef3b4035abf65566e9a8bfd642b4a2032729656e/) produces both problems. This is a POC to print the subject:

```
import lief
import sys

file_path = sys.argv[1] if len(sys.argv) > 1 else "example.exe"
binary = lief.parse(file_path)
for signature_index, signature in enumerate(binary.signatures):
        for signer in signature.signers:
                print(signer.cert.subject)
```
This code and this file will produce this output:
`'serialNumber=91140802MADALQC44B, ??=CN, ??=Private Organization, C=CN, ST=, O=, CN='`

**Expected behavior**
For this file should produce the following output:
`CN=运城市盐湖区风颜商贸有限公司, O=运城市盐湖区风颜商贸有限公司, ST=山西省, C=CN, Business Category=Private Organization, Jurisdiction Country=CN, Serial=91140802MADALQC44B`

**Environment (please complete the following information):**
 - System and Version : MacOS 15.5 (also observed from Ubuntu 20) 
 - Target format: PE
 - LIEF commit version: 0.16.6-d52c66d6

**Additional context**
Regarding the failure to convert OID strings, I checked the [OID to String table](https://github.com/lief-project/LIEF/blob/c678f6bdfb3f3199b36e1f8df25f7432a2725f53/src/PE/signature/OIDToString.cpp) and they are both represented. ( 2.5.4.15 and 1.3.6.1.4.1.311.60.2.1.3 , respectively.)

romainthomas:
Hi @Squiblydoo,

I can reproduce the issue for the non valid characters but how do you trigger the *The failure to convert some OID*? 

Please note that I'll check the effort to support non-roman characters but this is not in my short/mid terms priorities for LIEF

Squiblydoo:
I had the failure to convert some OID at the same time. 

[...]

PR #1226, opened 2025-06-15
Handle non-roman characters and extend OID To String support.

Squiblydoo:
Regarding this Issue: https://github.com/lief-project/LIEF/issues/1219

First issue:
Non-roman characters don't print because x509.cpp explicitly limits the printable characters to the range "32" to "127" 
https://github.com/lief-project/LIEF/blob/75e359bec0d70b6ae25e12c4c3ac54a32a622bdb/src/PE/signature/x509.cpp#L125

This pull request removes the upper limit. This works as expected, see example below.

Second issue:
Conversion of OID To String was dependent on "mbedtls_oid_get_attr_short_name" from mbedtls, in instances where a OID was not found in mbedtls, LIEF returned "??".

Instead of outputting "??", we use the `oid_to_string` method defined in `OIDToString.cpp` to supplement mbedtls's OID lookup table. If the OID is still not found, we return the OID Numeric string. 

Before the change:
`
serialNumber=91140802MADALQC44B, ??=CN, ??=Private Organization, C=CN, ST=, O=, CN=`

Result of DER parsing after change:
`serialNumber=91140802MADALQC44B, JURISDICTION_OF_INCORPORATION_C=CN, BUSINESS_CATEGORY=Private Organization, C=CN, ST=山西省, O=运城市盐湖区风颜商贸有限公司, CN=运城市盐湖区风颜商贸有限公司`

--------------------------
Please feel free to make any stylistic changes as desired, but it seems like this meets the objectives in a fairly straight forward way relying on your existing OIDToString mapping. 

Note: I compiled and tested locally against a small number of files. I didn't test it against malformed certificates that could have malicious content in the fields. (I don't know if that is a thing that exists or not.)

### DIFF
--- a/src/PE/signature/x509.cpp
+++ b/src/PE/signature/x509.cpp
@@ -103,13 +103,11 @@ int lief_mbedtls_x509_dn_gets( char *buf, size_t size, const mbedtls_x509_name *
         if( ret == 0 ) {
             ret = mbedtls_snprintf( p, n, "%s=", short_name );
         } else {
-            // Get the numeric OID string
+            // Get OID numeric string and return x509 friendly name or OID numeric string
             char oid_str[64] = {0};
             mbedtls_oid_get_numeric_string(oid_str, sizeof(oid_str), &name->oid);
-
-            // Use your friendly name mapping
-            const char* friendly = LIEF::PE::oid_to_string(oid_str);
-            ret = mbedtls_snprintf(p, n, "%s=", friendly);
+            const char* friendly_name = LIEF::PE::oid_to_string(oid_str);
+            ret = mbedtls_snprintf(p, n, "%s=", friendly_name);
         }
         MBEDTLS_X509_SAFE_SNPRINTF;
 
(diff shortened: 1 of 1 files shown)


src/PE/signature/OIDToString.cpp
```diff
@@ -2246,9 +2246,14 @@ const char* oid_to_string(const oid_t& oid) {
     { "1.3.6.1.4.1.311.10.3.28",          "szOID_PLATFORM_MANIFEST_BINARY_ID" },
     { "1.3.6.1.4.1.311.2.6.1",            "SPC_RELAXED_PE_MARKER_CHECK_OBJID" },
   };
-  auto   it  = oid_to_str.find(oid);
-  return it == oid_to_str.end() ? oid.c_str() : it->second;
+  auto it = oid_to_str.find(oid);
```
romainthomas, line 2249: What is the motivation behind this change?

romainthomas: I missed to read your comment before writing that but I still to get what you aim to do with this change

romainthomas: Ok I see what went wrong. `oid_to_string` is expecting a `const std::string&` but in your code you (implicitly) create the `std::string` in during the call. I'll fix that don't worry.

Squiblydoo:
Seems good to me. Thank you.

Issue #1215, opened 2025-05-20
Java Class files misidentified as Mach-O FAT files in `is_macho()`


wangmot:
**Describe the bug**
LIEF's `is_macho()` may be incorrectly identifying Java Class files as Mach-O FAT files. `is_macho()` checks the first 4 magic bytes of the file to determine if the file is a Mach-O file. But Java Class and Mach-O FAT files have the same 7 bytes in the beginning so by checking only the first 4 bytes, it's coming to the conclusion that Java Class files are Mach-O FAT files. 

**To Reproduce**
We ran it with a file called `HelloWorld.class`, which is a `compiled Java class data` file. This small error can be reproduced with any Java Class file and by giving it as an argument to the `is_macho()` function:

`lief.is_macho(file)`

**Additional context**
`is_macho()` calls `magic_from_stream()` to read the first 4 bytes.
Maybe it would be better to check the first 8 bytes to avoid this issue, or there could be other solutions as well. 

First 8 bytes of Java Class files: `ca fe ba be 00 00 00 37`
First 8 bytes of Mach-O FAT files: `ca fe ba be 00 00 00 02`

romainthomas:
You are right I'll see how to improve this check

PR #1260, opened 2025-10-31
differentiate between macho fat and java class files

wangmot:
This PR addresses the issue: https://github.com/lief-project/LIEF/issues/1215. 

Prevent Java class files from being misidentified as Mach-O FAT binaries by checking bytes 5-8. The first 4 bytes of java class and macho fat files are `CA FE BA BE`. The updated logic reads the first 8 bytes of the file and verifies whether the minor/major version falls within the valid Java class file range. If so, the file is not identified as a Mach-O FAT file. This change helps file type detection where Java class files were incorrectly identified as Mach-O FAT binaries.


romainthomas:
Sorry I made a mistake, I wanted to push a commit to the branch you created but somehow I failed.

Issue #1262, opened 2025-11-05
Mach-O parse_dyld_exports() hangs for certain binaries


trevor-e:
**Describe the bug**
I'm testing `lief` against some MachO binaries I've downloaded from the App Store (using `ipatool`) and I found a case where one of them causes a hang when parsing, my CPU gets pinned to 100% usage. Here is the symbolicated sample of the process, you can see it gets stuck somewhere with `parse_export_trie()`:

```
Call graph:
    2317 Thread_457403   DispatchQueue_1: com.apple.main-thread  (serial)
    + 2317 start  (in dyld) + 6076  [0x180e46b98]
    +   2317 Py_BytesMain  (in libpython3.13.dylib) + 40  [0x1017eaaec]
    +     2317 pymain_main  (in libpython3.13.dylib) + 468  [0x1017eaccc]
    +       2317 Py_RunMain  (in libpython3.13.dylib) + 1712  [0x1017eb538]
    +         2317 pymain_run_module  (in libpython3.13.dylib) + 232  [0x1017ebf88]
    +           2317 _PyEval_EvalFrameDefault  (in libpython3.13.dylib) + 17712  [0x10160aa44]
    +             2317 cfunction_vectorcall_FASTCALL_KEYWORDS.llvm.4622243580759790771  (in libpython3.13.dylib) + 88  [0x1018a25d4]
    +               2317 builtin_exec  (in libpython3.13.dylib) + 388  [0x1017087d8]
    +                 2317 PyEval_EvalCode  (in libpython3.13.dylib) + 132  [0x101708cd8]
    +                   2317 _PyEval_EvalFrameDefault  (in libpython3.13.dylib) + 17712  [0x10160aa44]
    +                     2317 cfunction_vectorcall_FASTCALL_KEYWORDS.llvm.4622243580759790771  (in libpython3.13.dylib) + 88  [0x1018a25d4]
    +                       2317 builtin_exec  (in libpython3.13.dylib) + 388  [0x1017087d8]
    +                         2317 PyEval_EvalCode  (in libpython3.13.dylib) + 132  [0x101708cd8]
    +                           2317 _PyEval_EvalFrameDefault  (in libpython3.13.dylib) + 60644  [0x1016151f8]
    +                             2317 slot_tp_call  (in libpython3.13.dylib) + 216  [0x10170b40c]
    +                               2317 _PyObject_Call_Prepend  (in libpython3.13.dylib) + 152  [0x10170b96c]
[...]

PR #1263, opened 2025-11-06
Memoize symbol names for export trie parsing

trevor-e:
`memoized_symbols_` was being populated in some code paths but not others, which for some binaries caused a gnarly O(n^2) lookup that would basically never succeed when parsing the export trie. I moved that logic into `parse_export_trie()` so it's always populated, rather than assuming it was populated before calling `parse_export_trie()`.

This fixes https://github.com/lief-project/LIEF/issues/1262 for me.

### DIFF
--- a/tests/macho/test_exports_trie.py
+++ b/tests/macho/test_exports_trie.py
@@ -1,7 +1,8 @@
 #!/usr/bin/env python
 import lief
 import subprocess
-from utils import get_sample, is_apple_m1, sign, chmod_exe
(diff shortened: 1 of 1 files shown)


romainthomas:
Thank you for raising the issue and creating a PR to address it. Would you mind to share a sample that trigger this bug so that I can reproduce it on my end and create a test for that?

trevor-e:
@romainthomas Awesome, yea removing the recursion guard is fine. Thanks for adding the test case, lgtm for merging!

Issue #470, opened 2020-09-27
Add Node.js Bindings/SDK


calebboyd:
**Is your feature request related to a problem? Please describe.**
It'd be great to have a Node.js SDK

**Describe the solution you'd like**
Generated or manually exposed Node APIs.
It looks like this project might be useful for generating bindings with little boilerplate for cmake projects
https://github.com/Geode-solutions/genepi

signadou:
We've created Node.js bindings for LIEF (for use with [tweakcc](https://github.com/Piebald-AI/tweakcc)): https://github.com/Piebald-AI/node-lief

PR #1269, opened 2025-11-20
Mention `node-lief` in the README and docs

signadou:
Thank you for this amazing library!  We're using it with great success in [tweakcc](https://github.com/Piebald-AI/tweakcc) to unpack and repack Bun executables on Windows, Linux, and macOS Intel/ARM.  In doing so we've needed to create Node.js bindings for LIEF [here](https://github.com/Piebald-AI/node-lief), and we've published them to [npm](https://www.npmjs.com/package/node-lief).

I'd like to add it to the README so that users can see if it they have a need for it (issue #470 is an example of that).


signadou:
@romainthomas Good point, I forgot all about tests.  I've added 257 tests with 97.4% code coverage (see the badge on the README).

PR #1278, opened 2025-12-10
Add missing segments in X86_64 coredump parser

llaubin:
Hi o/

This small PR just adds the missing segments when loading an X86_64 elf coredump. This is mainly useful to deal with `fs_base`  if you want to start some emulation from the coredump of a binary after a breakpoint ;-)


romainthomas:
The fix is part of the [`0.17.5`](https://github.com/lief-project/LIEF/releases/tag/0.17.5) release

Issue #1284, opened 2026-01-04
PE Builder Corrupts the Output Binary


szescxz:
**Describe the bug**
I am patching [the Steam version of Industry Giant 2](https://store.steampowered.com/app/271360/). Access to the binary in question should have sent to contact@lief.re with a subject same as the title of this issue (let me know if this is not the case), and you can verify its SHA-1 hash against [the one listed on SteamDB](https://steamdb.info/depot/271361/?show_hashes).
Despite the binary is protected by the Steam DRM, it is not packed and therefore enabling the possibility to patch it; however, even letting LIEF to rebuild the binary without any modifications will result in a `EXCEPTION_ACCESS_VIOLATION` crash upon launching.
If I set [`force_relocating`](https://lief.re/doc/stable/formats/pe/python.html#lief.PE.Builder.config_t.force_relocating) to `True`, then the rebuilt binary works.

**To Reproduce**
```python
import lief
binary = lief.parse("ig2_AddOn.exe")
config = lief.PE.Builder.config_t()
# config.force_relocating = True # <--- Uncomment this line works
binary.write("patched.exe", config)
```

**Expected behavior**
Binary behavior not altered even when `force_relocating` is set to `False`

**Environment**
 - System and Version: Windows 11
 - Target format: PE
 - LIEF commit version: `0.17.1-fe54643f` as well as `0.17.2-aa2b617f`

**Additional context**
Running [`VBinDiff`](https://www.cjmweb.net/vbindiff/) on both binaries shows that LIEF somehow copied contents starting from file offset 0x331E00 to 0x0B4400.
<details><summary><code>VBinDiff</code> output</summary>
<pre><code>ig2_AddOn.exe
000B 43C0: 10 8B 43 04 C7 45 FC 00  00 00 00 89 46 14 FF 15  .婥.荅? ...塅..
000B 43D0: 24 D7 6C 00 89 46 18 8B  45 10 89 46 1C 8B 45 14  $譴.塅.?E.塅.婨.
000B 43E0: 89 46 20 E8 08 C3 15 00  84 C0 74 3A E8 EF AD F5  塅 ??. 劺t:栾  .
000B 43F0: FF 84 C0 74 31 8B 0F 8B  03 8B 57 04 03 C1 89 4D  劺t1??.媁..翂M
000B 4400: E0 8B 4B 04 03 CA 89 55  E4 89 4D EC 8D 55 E0 8B  鄫K..蕢U 鋲M鞃U鄫
[...]

PR #1296, opened 2026-02-08
Fix address type

Immortalety:
The `build_tls` function uses the wrong address type when calling `patch_address`, which results in an error: `Can't find section with the rva: 0xfffffffec043e788` (large negative number)


romainthomas:
Thank you. This is an error introduced when fixing #1284

PR #1301, opened 2026-02-12
Fix `MachO::Binary::virtual_address_to_offset()` for non-file-backed segment

### LINKED ISSUE #1299: `LIEF::MachO::Binary::virtual_address_to_offset()` returns incorrect result if VA lies in non-file-backed segment
**Describe the bug**
`LIEF::MachO::Binary::virtual_address_to_offset()` returns an incorrect result if the provided virtual address lies in a non-file-backed segment, e.g. a `__DATA` segment consisting only of `S_ZEROFILL` sections.

**To Reproduce**
It turns out that the `MachO/do_add.bin` already serves as a demonstrator.  This executable file has a `__DATA` segment comprised of only the `_bss` section (which is `S_ZEROFILL`).  Calling `LIEF::MachO::Binary::virtual_address_to_offset(0x100008000)` (which points to start of the `__DATA` segment), returns 0.

**Expected behavior**
In such a case, `virtual_address_to_offset()` should probably return `lief_errors::conversion_error` instead.

 - System and Version : N/A
 - Target format Mach-O
 - LIEF commit version: reproduces on `main` branch

**Additional context**

jalopezg-git:
`LIEF::MachO::Binary::virtual_address_to_offset()` returns an incorrect result if the provided virtual address lies in a non-file-backed segment, e.g. a `__DATA` segment consisting only of `S_ZEROFILL` sections.

It turns out that the `MachO/do_add.bin` already serves as a demonstrator.  This executable file has a `__DATA` segment comprised of only the `_bss` section (which is `S_ZEROFILL`).  E.g. calling `LIEF::MachO::Binary::virtual_address_to_offset(0x100008008)` (which points to start of the `__DATA` segment), returns 8.  Instead, `virtual_address_to_offset()` should return `lief_errors::conversion_error`.

This pull request fixes #1299.

### DIFF
--- a/src/MachO/Binary.cpp
+++ b/src/MachO/Binary.cpp
@@ -1842,7 +1842,9 @@ std::vector<uint8_t> Binary::raw() {
 
 result<uint64_t> Binary::virtual_address_to_offset(uint64_t virtual_address) const {
   const SegmentCommand* segment = segment_from_virtual_address(virtual_address);
-  if (segment == nullptr) {
+  // file_size() is 0 if segment is not file-backed, e.g. a `__DATA` segment that only has
+  // a ZEROFILL section.  In such case, there is no file offset available.
+  if (segment == nullptr || segment->file_size() == 0) {
     return make_error_code(lief_errors::conversion_error);
   }
   const uint64_t base_address = segment->virtual_address() - segment->file_offset();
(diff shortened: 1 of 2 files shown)


romainthomas:
This makes completely sense. Thank you!

PR #1332, opened 2026-04-25
Update rust crates dependencies

BraveChicken1:
This PR updates the dependencies of the several rust crates.

It bumps the versions of dependencies, removes unused dependencies and moves the tempfile dependency to the dev-dependencies, since it's only used by the tests. It also changes the `elf_tests.rs` test, by ensuring it does not use the reserved `gen` keyword anymore.
As a result of this the MSRV is bumped to 1.85.0.

By bumping the dependencies, we can benefit of the stability improvements of these dependencies, and more important several vulnerability fixes to these dependencies.
Dependents of LIEF benefit of this, because the overal security and stability of the bindings improve and their dependency tree does not need to contain older package versions anymore. This then also results in improved security and stability of the dependents.

Please note that miette, zip and cxx were not updated to the latest versions, because of compatibility issues with autocxx (miette and cxx) and because I don't want to bump the MSRV too far (zip 8.5.1 requires MSRV 1.88.0, while 1.85.0 is the current standard for big Rust projects).

To support upgrading miette and cxx, we would need to bump the dependencies of autocxx too. The autocxx project seems to be a bit inactive at the moment, but I will try to get the changes into the latest version of autocxx. This will then also contain the bump of the `aquamarine` dependency, which you have changed on your own fork of autocxx. I will open a follow up PR once autocxx is upgraded.

LIEF still builds and works after these changes, I tested this on several different types of machines. I also read through the changelogs of the different dependencies to ensure there were no breaking changes for LIEF.

Some small final notices:
- These changes open the path to upgrading the used rust edition of LIEF to 2024, I'm willing to look into this in a future PR.
- The main branch contains several commits and PRs (including my PRs from several weeks ago) that were not part of the latest release. These commits seem to contain several fixes and improvements that would be nice to have in the next release.


romainthomas:
Thank you. 

Two questions:

1. Is this AI-generated? If so, in which extent?
2. Did you test on Windows?

BraveChicken1:
No, I did not use AI to generate any of this. I read through the changelogs and crates.io pages myself and then upgraded the dependencies to the most recent, still compatible version. I know the diff is very big, but this is mainly because the lock files were regenerated by cargo.

I did test if lief still builds and runs on Windows, and this does indeed still work.

romainthomas:
Ok thank you. I'm not against bumping the MSRV but I'll need to assess the impact on the build pipeline

BraveChicken1:
Yes, I understand that bumping the MSRV has some impact on the library, but I think bumping it now is a good moment to future proof LIEF.

Let me know if there is anything I can help with.

romainthomas:
Just to let you know, any bump of MSRV will come after the `1.0.0` release

BraveChicken1:
When do you plan on doing the 1.0.0 release? If this is still a long time, would it maybe be nice if I submit a PR with fixes and bumps that do not yet require a MSRV bump?

romainthomas:
Yes I would be ok with this kind of update

romainthomas:
As you can see in the associated commit, I did a global refactoring of the low level Rust bindings 
which was worth raising the MSRV. Feel free to provide feedback on this new layout.

BraveChicken1:
Thank you for this refactor, it looks really good! My only feedback would be to be cautious with specifying dependency versions only with `major.minor`, since this could accidentally pull in a newer version which is not compatible. Do you already know when this refactored version will be released?

romainthomas:
Ok thank you. I'll push another update to fix `major + minor`. 

> Do you already know when this refactored version will be released?

For the v1.0.0 which I plan to release around July

PR #1347, opened 2026-05-19
doc/sphinx/tutorials/12_elf_coredump.rst: Fix typo

isodude:
Tried to use the code from the docs, according to https://lief.re/doc/latest/formats/elf/python.html#lief.ELF.Note it should be CORE_FILE rather than CORE.FILE though.

Happens to the best of us.


romainthomas:
Thank you

PR #1357, opened 2026-06-29
fix: c++20 ambiguous namespace fmt

Hennzau:
In addition to https://github.com/lief-project/LIEF/pull/1339 this commit fix some PE files. LIEF can now be built with CXX_STANDARD set to 20.

As C++20 is not exactly officially supported, not sure if you want this pr @romainthomas. But I needed it for my current work so here are some fixes


romainthomas:
I'm good with these changes. I think the cleanest way would be to completely remove the `using namespace fmt;` are use `fmt::format`

PR #1364, opened 2026-07-26
Optimize ELF data insertion when appending content

quadratic-bit:
We have a use case for LIEF's ELF engine that involves appending a large number of segments to an ELF file. Adding a loaded section to an `ET_EXEC` binary also creates a segment. Several operations on this path previously performed linear searches over the binary layout or `DataHandler` nodes for each addition, causing the total cost to grow superlinearly as the number of sections and segments increased.

Profiling with `perf`, I optimized several sources of quadratic behavior, along with some significant linear-time costs.

### Performance

The benchmark repeatedly adds loaded sections to an ELF binary and serializes the result in memory with `Binary::raw()`.
The plots below compare the original and optimized implementations across small, medium, and large segment-count ranges (I'll provide any additional details if needed)
<p align="center">
Small benchmark comparison
Medium benchmark comparison
Large benchmark comparison
</p>

While the primary target is workloads involving more than 10,000 segments, the changes also improve general case of 10-100 segments.

### Build
Built with
```
cmake -S . -B build
cd build
make
```

### DIFF
--- a/api/python/src/ELF/objects/pyBinary.cpp
+++ b/api/python/src/ELF/objects/pyBinary.cpp
@@ -853,6 +853,10 @@ void create<Binary>(nb::module_& m) {
       "array_tag"_a
     )
 
+    .def("reserve_segments",
+        &Binary::reserve_segments,
+        "count"_a)
+
     .def("find_version_requirement",
       nb::overload_cast<const std::string&>(&Binary::find_version_requirement),
       R"doc(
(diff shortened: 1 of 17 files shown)


romainthomas:
> We have a use case for LIEF's ELF engine that involves appending a large number of segments

I'm curious about this use case :)

quadratic-bit:
The use case I mentioned involves generating checkpoints similar to [ELFies](http://heirman.net/papers/patil2021elfies.pdf). We use our own binary files generator, and the number of segments is usually quite big.

I submitted the changes because they show consistent improvements even for low number of segments. And on 10k+ segments quadratic behavior overtakes (as seen on the third pic), so the difference in performance is very noticeable.

I would also argue about the assigned labels. I used AI assistance only for generating the test file for cc17f18, as the cache may be a bit fragile, and the test would prevent someone unknowingly violate the invariance. Everything else (including other code and PR description) is hand-written.

romainthomas:
Is this PR sincerely 100% hand-made written except the tests?

quadratic-bit:
This PR is sincerely written by my hand in a span of a week, except for the test-file, for which I brainstormed edge cases.

PR #1367, opened 2026-07-28
Fix MIPS n64 ELF relocation parsing

plusls:
MIPS n64 (ELFCLASS64 + EM_MIPS) uses a non-standard on-disk `r_info` layout: `r_info = r_sym[4] | r_ssym[1] | r_type3[1] | r_type2[1] | r_type[1]` instead of the generic ELF64 `(sym << 32) | type`. LIEF decoded all ELF64 relocations with the generic mask, which is wrong for MIPS n64:

- Little-endian: `r_sym` is in the low 32 bits, so `r_info >> 32` produces garbage symbol indices. This caused the AUTO dynsym count to explode (196 instead of 10) and relocation types to decode as UNKNOWN.

- Big-endian: `r_info & 0xffffffff` drags r_ssym/r_type3/r_type2 into the type field. Compound relocations (e.g. R_MIPS_GPREL16 with non-zero Type2/Type3) decode as garbage; simple ones match by accident.

The MIPS n64 encode/decode is now isolated in an internal header `src/ELF/reloc_utils.hpp`, keeping the public Relocation class format-agnostic (same pattern as the existing Android SLEB128 handling at the Parser/Builder boundary).

The judgment key `arch == MIPS && class == ELF64` matches binutils `ABI_64_P` macro. n32/o32 are ELFCLASS32 and handled by the Elf32 code path at compile time, so they cannot trigger the MIPS n64 branch.

Non-MIPS architectures: zero change in decode/encode paths.


TaylorSasser:
Correct me if I'm wrong, but I think Relocation::resolve need an additional case for MIPS_REL32 so this can be used in the Python/Rust bindings.

Possibly something like this?
```    
case TYPE::MIPS_REL32:
      return (info() == 0 ? B : S) + Q();
```

PR #1368, opened 2026-08-06
MachO: adjust __init_offsets when add_section shifts __TEXT

### LINKED ISSUE #1229: Shifting the text segment invalidates load address relative addresses in MachO binaries
**Describe the bug**
`LIEF::MachO::Binary::extend` is used by `LIEF::MachO::Binary::add_section` to shift existing sections if we attempt to inject a new section into to the `__TEXT` segment in order to keep relative offsets intact. However, some sections are containing pointers that are stored as load address based relative offsets (eg.: `INIT_FUNC_OFFSETS`) which would not be valid anymore.

**To Reproduce**
Inject a non zero sized section (with 0 offset) to the `__TEXT` segment of a binary which has a constructor method (`__attribute__((constructor))`). Running the application should result in a crash as the constructor function's offset is likely pointing to an address that no longer contains valid instructions.

**Expected behavior**
Shifting of the segments should update relative pointers akin to  how the load commands are updated.

 - System and Version : MacOS
 - Target format: `Mach-O`
 - LIEF commit version: `0.16.6`

**Additional context**
I haven't checked other section types which might also susceptible to this issue, as i'm not sure how other load address based pointers (eg.: MOD_INIT_FUNC_POINTERS) get identified/finalised by dyld.

je-empty:
`add_section` into `__TEXT` shifts the following content via `Binary::shift`, but `shift` doesn't update the image-base offsets stored in `__TEXT,__init_offsets` (`S_INIT_FUNC_OFFSETS`). They keep pointing at the old location, so dyld jumps into the wrong place and the binary crashes at launch (SIGILL). The fix bumps each entry past the shift point.

```python
b = lief.MachO.parse("app").at(0)   # app has a constructor -> __init_offsets
b.add_section(lief.MachO.Section.create("__mysec", [0x41] * 0x4000))  # into __TEXT
b.write("out")   # SIGILLs on launch without the fix, runs with it
```

The constructor offset stays `0x4b0` while the code moves to `0x44b0`; a section too small to force a shift is unaffected. Verified on macOS arm64 and an iPhone Air (iOS 26.6). Closes #1229.

### DIFF
--- a/src/MachO/Binary.cpp
+++ b/src/MachO/Binary.cpp
@@ -1005,6 +1005,24 @@ ok_error_t Binary::shift(size_t value) {
       }
     }
   }
+  // The shift moved the code in __TEXT but not the image-base offsets in
+  // __TEXT,__init_offsets, so bump the ones past the shift point.
+  for (Section* section : sections_) {
+    if (section->type() != Section::TYPE::INIT_FUNC_OFFSETS) {
+      continue;
+    }
+    span<uint8_t> content = section->content();
+    for (size_t i = 0; i + sizeof(uint32_t) <= content.size();
+         i += sizeof(uint32_t)) {
+      uint32_t init_off = 0;
+      std::memcpy(&init_off, content.data() + i, sizeof(init_off));
+      if (init_off >= loadcommands_end) {
+        init_off += static_cast<uint32_t>(value);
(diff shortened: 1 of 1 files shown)


je-empty:
Superseded by https://github.com/lief-project/LIEF/pull/1369 — same fix, clean single-commit history.

PR #1376, opened 2026-08-30
Require MbedTLS 4.0.0 exactly

jeongseok-meta:
## Summary

- require MbedTLS 4.0.0 exactly when `LIEF_OPT_MBEDTLS_EXTERNAL` is enabled
- propagate the same exact dependency through the installed LIEF CMake package

LIEF currently checks `MBEDTLS_VERSION_NUMBER == 0x04000000` and relies on MbedTLS 4.0 private PSA APIs, but `find_package(MbedTLS 4.0.0 REQUIRED)` also accepts later 4.x releases. This allowed MbedTLS 4.2.0 through configuration and then failed during compilation with the version assertion, changed private members, and a removed PSA header, as seen in the [conda-forge build](https://github.com/conda-forge/lief-feedstock/actions/runs/33276466411).

## Test plan

- Configure a minimal CMake consumer against MbedTLS 4.0.0: succeeds.
- Configure the same consumer against MbedTLS 4.2.0: rejected at `find_package` as not exactly 4.0.0.
- `git diff --check`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -144,7 +144,7 @@ if(NOT LIEF_OPT_MBEDTLS_EXTERNAL)
     -DTF_PSA_CRYPTO_USER_CONFIG_FILE="${CMAKE_CURRENT_SOURCE_DIR}/config/mbedtls/psa_config.h"
   )
 else()
-  find_package(MbedTLS 4.0.0 REQUIRED)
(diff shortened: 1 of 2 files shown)


romainthomas:
Thank you. I cherry-picked on the [release/1.0.x](https://github.com/lief-project/LIEF/commits/release/1.0.x) branch

